PR #8, opened 2015-11-19
Add rpad to pad a buffer with trailing zeroes

axic:
Useful for handling the ```bytes``` type in the ABI. I'm using it [here](https://github.com/axic/ethereumjs-abi).


wanderer:
thanks @axic

PR #12, opened 2015-12-05
fix private to public is in v2.3.2

ethers:


wanderer:
thanks @ethers I already had this one bumped and tagged but must have forgot to push

ethers:
:)

PR #27, opened 2016-03-07
Ecrecover only accepting v in {27,28}

MrChico:


wanderer:
thanks @MrChico published as 4.0.2 !

PR #30, opened 2016-03-21
Implement verifyPrivate for verifying a private key

axic:
Not sure about naming, could also be:
- `privateKeyVerify`
- `validatePrivate` / `validatePrivateKey` / `privateKeyValidate`
- `isValidPrivate`/`isValidPrivateKey`

### DIFF
--- a/index.js
+++ b/index.js
@@ -296,6 +296,16 @@ exports.rlphash = function (a) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


axic:
`publicKeyCreate` does verification also, but I think it is useful to have it available directly for various tools & wallets.

wanderer:
`isValidPrivate` seem appropriate

PR #32, opened 2016-04-14
Introduce isValidPublic()

axic:


index.js
```diff
@@ -306,6 +306,27 @@ exports.isValidPrivate = function (privateKey) {
...
+ * and the requirements of Ethereum.
+ * @method isValidPublic
+ * @param {Buffer} publicKey The two points of an uncompressed key, unless sanitize is enabled
+ * @param {Boolean} [sanitize=false] Accept public keys in other formats
+ * @return {Boolean}
+ */
+exports.isValidPublic = function(publicKey, sanitize) {
+  if (publicKey.length === 64) {
+    // Convert to SEC1 for secp256k1
+    return secp256k1.publicKeyVerify(Buffer.concat([ new Buffer('04', 'hex'), publicKey ]))
```
axic: @fanatid can you review this?

SEC1 2.0 (http://www.secg.org/sec1-v2.pdf) only lists 04 for uncompressed keys, but the secp256k1 code has 06 and 07 too.

index.js
```diff
@@ -306,6 +306,27 @@ exports.isValidPrivate = function (privateKey) {
...
+ * and the requirements of Ethereum.
+ * @method isValidPublic
+ * @param {Buffer} publicKey The two points of an uncompressed key, unless sanitize is enabled
+ * @param {Boolean} [sanitize=false] Accept public keys in other formats
+ * @return {Boolean}
+ */
+exports.isValidPublic = function (publicKey, sanitize) {
+  if (publicKey.length === 64) {
+    // Convert to SEC1 for secp256k1
+    return secp256k1.publicKeyVerify(Buffer.concat([ new Buffer('04', 'hex'), publicKey ]))
```
axic: @fanatid can you review this?

SEC1 2.0 (http://www.secg.org/sec1-v2.pdf) only lists `04` for uncompressed keys, but the secp256k1 code has `06` and `07` too.

fanatid: @axis looks good
you can write optimized version:
```js
var _uncompressedPublicKey = new Buffer(65)
_uncompressedPublicKey[0] = 0x04
exports.isValidPublic = function (publicKey, sanitize) {
  if (publicKey.length === 64) {
    publicKey.copy(_uncompressedPublicKey, 1)
    return secp256k1.publicKeyVerify(_uncompressedPublicKey)
  }
  ...
}
```

fanatid: 0x06 and 0x07 checks that y point is odd or not, like 0x02 and 0x03 for compressed key

axic: The Buffer.copy is nice, but don't want to pollute the context with random data. Do you think moving the whole into the if would still be faster than Buffer.contact? I think it would.

So 0x06 means it is odd and 0x07 means it is even (or the other way around?); while 0x04 doesn't do that check?

fanatid:
0x07 for odd -- https://github.com/bitcoin/secp256k1/blob/7b549b1abc06fe1c640014603346b85c8bc83e0b/src/eckey_impl.h#L27

You don't pollute context with `_uncompressedPublicKey`, but actually it's not good solution because use global variables.. I think stay with concat will be better:
```js
return secp256k1.publicKeyVerify(Buffer.concat([ new Buffer([0x04]), publicKey ]))
```

index.js
fanatid: 0x07 for odd -- https://github.com/bitcoin/secp256k1/blob/7b549b1abc06fe1c640014603346b85c8bc83e0b/src/eckey_impl.h#L27

You don't pollute context with `_uncompressedPublicKey`, but actually it's not good solution because use global variables.. I think stay with concat will be better:
```js
return secp256k1.publicKeyVerify(Buffer.concat([ new Buffer([0x04]), publicKey ]))
```

axic: Thanks for all the help!  Used this last version of concat with `new Buffer([4])`, should be good enough.

PR #33, opened 2016-04-14
Implement isValidSignature()

axic:
No tests yet, WIP.

### DIFF
--- a/index.js
+++ b/index.js
@@ -584,7 +584,7 @@ exports.padToEven = function (a) {
  * @param {Buffer} v
  * @param {Buffer} r
  * @param {Buffer} s
- * @param {Boolean} [homestead=false]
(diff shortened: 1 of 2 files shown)


axic:
@wanderer should `ecrecover()` have a `homestead` flag too?

wanderer:
@axic ecover should actually be ok, If someone can find a tx that that has high r values in the chain then we will have to add it, but I don't think that is case.

fanatid:
`isVlidSignature` is not the best name, maybe `isValidSignatureNumbers`?

axic:
@fanatid I agree. Would `isValidSignatureParameters` work?

fanatid:
@axic  I don't know, anything except `isValidSignature` :smile:

PR #34, opened 2016-04-18
Implement importPublic()

axic:
Method to convert any public key to an Ethereum public key.

### DIFF
--- a/index.js
+++ b/index.js
@@ -357,6 +357,20 @@ var privateToPublic = exports.privateToPublic = function (privateKey) {
 }
 
 /**
+ * Converts a public key to the Ethereum format.
+ * @method importPublic
+ * @param {Buffer} publicKey
+ * @return {Buffer}
+ */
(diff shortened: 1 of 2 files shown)


wanderer:
@axic it looks like we might have enough functions for keys that we could split all the key stuff out into its own file or even module. What do you think?

axic:
@wanderer I was thinking the same today to splitting into `ethereumjs-crypto`. It is hard to split it correctly though:
a) everything which requires secp256k1?
b) everything which is address/key related?

Most of the ethereumjs packages will need the crypto part too, though other users could avoid using that if they only need buffer utils (that's left in utils after crypto).

wanderer:
which out of `a` or `b` is the largest subset?  I think its `b`.  Then we could split ethereumjs-util into two files `key-stuff.js` and `buffer-stuff.js`. `index.js` would still export all the function in both the files but users could decided to only include one of the subfiles. Which would be good for shaving off bits for browsers

PR #37, opened 2016-04-28
Removing excessive secp256k1.publicKeyConvert

fanatid:
secp256k1-node v3 has option `compressed` for `publicKeyCreate`
https://github.com/cryptocoinjs/secp256k1-node/blob/2e3ee65a90d294343001e8b03bcf146ba9332ddb/API.md#publickeycreatebuffer-privatekey--boolean-compressed--true---buffer


wanderer:
awesome thanks @fanatid !

PR #39, opened 2016-05-17
Use create-hash

fanatid:


wanderer:
why do you want create-hash?

fanatid:
create-hash is a part of crypto-browserify (but in node this package uses node crypto)
with this browserify bundle will have smaller size

wanderer:
ok cool

PR #41, opened 2016-06-09
Make v an int in ecrecover()

axic:
This is a breaking change. It would be nice to have it merged in the future.


wanderer:
the reason it is this way is because the signuatures in ethereumjs-tx and and ethereumjs-vm are always a buffer. And it make the API more simple and unified

axic:
Is `v` only a Buffer in ethereumjs-tx because of the mess of RLP?

wanderer:
yeah, there is no types in RLP. so everything always comes out a buffer

wanderer:
ok incoming breaking change

PR #44, opened 2016-07-06
Feature - eth_sign signature method (different from tx sig)

Denton-L:
As described in ethereumjs/testrpc#103 and ethereumjs/testrpc#104, the `v` value returned for `ecsign` is inconsistent between Geth and testrpc nodes. In Geth, `v` is always `00` or `01`. In testrpc, `v` is always `1a`  (27) or `1b` (28). As a result, any dapps which use `ecsign` will have inconsistent signature values between the two nodes. Note that Geth should probably the "authority" for this.

I'm not sure whether this patch should be included in this repository or in testrpc. I feel like it should be included in testrpc because it has the potential to break less repositories than ethereumjs-util. Still, I'm making a pull request here for posterity.

close ethereumjs/testrpc#104

cc @tcoulter

### DIFF
--- a/index.js
+++ b/index.js
@@ -383,7 +383,7 @@ exports.ecsign = function (msgHash, privateKey) {
   var ret = {}
   ret.r = sig.signature.slice(0, 32)
   ret.s = sig.signature.slice(32, 64)
-  ret.v = sig.recovery + 27
+  ret.v = sig.recovery
   return ret
 }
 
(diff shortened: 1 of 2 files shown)


kumavis:
@Denton-L I updated [the eth rpc wiki](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign )

wanderer:
This would be a breaking change. The reason it exists like this is because that is how it is specified in the Yellow paper.

Denton-L:
So does that mean that Geth is broken?

wanderer:
No, I just think the RCP spec is different

Denton-L:
So, this change belongs in testrpc?

wanderer:
@Denton-L yes I think so

wanderer:
I will close for now. But feel free to reopen.

kumavis:
So tx's are signed using this method, eh

kumavis:
@wanderer it would be nice to have an `eth_sign` signature method that doesn't add the magic number, so module consumers can just use that. reopening...

kumavis:
Whoops - meant to re-open the issue, we should indeed close this PR

PR #46, opened 2016-08-03
changed pub key validation to sanitize when the key length is not 64 or explicitly requested

wanderer:
@axic is this safe behavor? Do we need to `assert` after the sanitization?


axic:
The name `sanitize` might be misleading. it should be called `allow-public-keys-from-broken-tools`. It enforces that only proper Ethereum public keys can be passed in. When turned on, it **tries** to convert them to a proper public key.

I think it should be left as it to increase visibility of the difference of an Ethereum public key and a SEC1 one.

wanderer:
ok

PR #47, opened 2016-08-08
Update wrong variable name sha3() method

tad88dev:
```javascript
/**
 * Creates SHA-3 hash of the input
 * @method sha3
 * @param {Buffer|Array|String|Number} a the input data
 * @param {Number} [bytes=256] the SHA width
 * @return {Buffer}
 */
exports.sha3 = function (a, bytes) {
  a = exports.toBuffer(a)
  if (!bytes) bytes = 256

  var h = new SHA3(bytes)
  if (a) {
    h.update(a)
  }
  return new Buffer(h.digest('hex'), 'hex')
}
```
I don't think that is bytes.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -373,7 +373,7 @@ Creates SHA-3 hash of the input
 **Parameters**
 
 -   `a` **Buffer or Array or String or Number** the input data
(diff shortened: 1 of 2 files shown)


fanatid:
ACK

wanderer:
awesome thank @tad88dev

Issue #40, opened 2016-06-05
Property alias does not work for setting with initial parameters


kumavis:
doesnt

``` js
tx = new Tx({ gas: 0x1234 })
```

works

``` js
tx = new Tx()
tx.gas = 0x1234
```

axic:
Have you found a solution to this?

axic:
@kumavis @wanderer is this solved?

wanderer:
i don't think so

kumavis:
rejoice!

PR #55, opened 2016-11-07
Fix property alias on initialization

fanatid:
Fix for https://github.com/ethereumjs/ethereumjs-util/issues/40 and https://github.com/ethereumjs/ethereumjs-tx/issues/22


wanderer:
super! thanks @fanatid . I also add you to the ethereumjs org

PR #58, opened 2016-11-09
Pad rpcSig output

### LINKED ISSUE #57: Padding zeroes issue with toRpcSig function
Sometimes resulting signature has values for r and s lower than 64 bytes.

At this moment function used to return signature works this way:
```
exports.toRpcSig = function (v, r, s) {
  // geth (and the RPC eth_sign method) uses the 65 byte format used by Bitcoin
  // FIXME: this might change in the future - https://github.com/ethereum/go-ethereum/issues/2053
  return exports.bufferToHex(Buffer.concat([ r, s, exports.toBuffer(v - 27) ]))
}
```
when r or v has less than 64 bytes, resulting concatenated string has a wrong length and value.

axic:
Fixes #57.

### DIFF
--- a/index.js
+++ b/index.js
@@ -391,6 +391,11 @@ exports.ecrecover = function (msgHash, v, r, s) {
  * @return {String} sig
  */
 exports.toRpcSig = function (v, r, s) {
+  // NOTE: with potential introduction of chainId this might need to be updated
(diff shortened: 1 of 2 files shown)


fanatid:
utACK

wanderer:
please adds tests

PR #59, opened 2016-11-10
Update Buffer creation

fanatid:
`new Buffer` is deprecated since node v6
is major change?


axic:
I think we still need to support pre-6 in util at least (see the Travis list of node versions).

fanatid:
@axic node v4 already support `alloc/allocUnsafe/from`, 0.12 will be not supported from 1 Jan 2017 AFAIR

wanderer:
lgtm

PR #62, opened 2016-11-29
Remove zero padding in intToHex

vethan:
The padded zeros when the value contains an uneven amount of digits causes a discrepency between how testrpc works, and how a true ethereum network works (e.g, the real ethereum network would return 0x1 for eth_getTransactionCount, whereas testrpc will return 0x01.)  

For quantities, testrpc behavior is against the [json rpc specification](https://github.com/ethereum/wiki/wiki/JSON-RPC#hex-value-encoding)

### DIFF
--- a/index.js
+++ b/index.js
@@ -181,7 +181,7 @@ exports.intToHex = function (i) {
  */
 exports.intToBuffer = function (i) {
   var hex = exports.intToHex(i)
-  return Buffer.from(hex.slice(2), 'hex')
+  return Buffer.from(exports.padToEven(hex.slice(2)), 'hex')
 }
 
 /**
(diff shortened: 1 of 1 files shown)


vethan:
Temporarily closing PR while work on fixing broader effects

axic:
@vethan have you tried this change with ethereumjs-tx and ethereumjs-vm?

vethan:
@axic Haven't checked ethereumjs-tx or ethereumjs-vm! I was running some more extensive tests this weekend, and unfortunatey this change wont fully fix the bigger issue.  Need to hunt down the bigger fix but not looking like I'll have time for a couple of weeks

wanderer:
im going to close this for now, please move the RP to https://github.com/ethjs/ethjs-util

PR #64, opened 2016-12-11
Use keccak instead keccakjs

fanatid:
keccak package -- https://github.com/cryptocoinjs/keccak


wanderer:
lgtm. the new keccak module is looking nice :)

fanatid:
Updated keccak version.. prebuild binaries added :)

PR #66, opened 2016-12-21
Name field is mandatory in npm >3

zmanian:
The name field is needed on all contributors in npm >3. Install fails without this fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
       "additions": 26,
       "deletions": 2
     },
-    {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,7 +38,8 @@
       "additions": 26,
       "deletions": 2
     },
-    {
+    { 
+      "name":"Mr. Chico"
```
wanderer: you are missing a ',' there

PR #67, opened 2017-01-20
Add helpers for new eth_sign behavior

yusefnapora:
This adds some helpers to deal with the recent changes to the `eth_sign` RPC call, as discussed in https://github.com/ethereumjs/testrpc/pull/244

The `hashForEthSign` function returns the prefixed and hashed input to the `ecsign` method for a given message, and is used by the `ethSign` function to produce a signature that matches the `eth_sign` output for a message.

There's also an `ethRecover` function, which just runs `ecrecover` on the output of `hashForEthSign`.

I'm not crazy about these function names, so feel free to suggest alternatives :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -367,7 +367,7 @@ exports.ecsign = function (msgHash, privateKey) {
  * @param message
  * @returns {Buffer} hash
  */
-exports.hashForEthSign = function (message) {
+exports.hashPersonalMessage = function (message) {
   var prefix = exports.toBuffer('\u0019Ethereum Signed Message:\n' + message.length.toString())
   return exports.sha3(Buffer.concat([prefix, message]))
 }
--- a/index.js
+++ b/index.js
@@ -373,19 +373,6 @@ exports.hashPersonalMessage = function (message) {
 }
 
 /**
- * ECDSA sign `message` using the same algorithm as the `eth_sign` RPC method.
- * Prefixes message with fixed header + `message.length.toString()`, then
- * signs the Keccak256 hash of the result.
- * @param {Buffer} message
- * @param {Buffer} privateKey
- * @return {Object} signature
- */
-exports.ethSign = function (message, privateKey) {
-  var msgHash = exports.hashForEthSign(message)
-  return exports.ecsign(msgHash, privateKey)
-}
-
-/**
  * ECDSA public key recovery from signature
  * @param {Buffer} msgHash
  * @param {Number} v
@@ -401,19 +388,6 @@ exports.ecrecover = function (msgHash, v, r, s) {
   }
   var senderPubKey = secp256k1.recover(msgHash, signature, recovery)
   return secp256k1.publicKeyConvert(senderPubKey, false).slice(1)
-}
-
-/**
- * ECDSA public key recovery from a message signed with the `eth_sign` RPC call.
- * @param {Buffer} message
- * @param {Number} v
- * @param {Buffer} r
- * @param {Buffer} s
- * @returns {Buffer} publicKey
- */
-exports.ethRecover = function (message, v, r, s) {
-  var msgHash = exports.hashForEthSign(message)
-  return exports.ecrecover(msgHash, v, r, s)
 }
 
 /**
--- a/index.js
+++ b/index.js
@@ -360,6 +360,32 @@ exports.ecsign = function (msgHash, privateKey) {
 }
 
 /**
+ * Returns the keccak-256 hash of `message`, prefixed with the header used by the `eth_sign` RPC call.
+ * The output of this function can be fed into `ecsign` to produce the same signature as the `eth_sign`
+ * call for a given `message`, or fed to `ecrecover` along with a signature to recover the public key
+ * used to produce the signature.
+ * @param message
+ * @returns {Buffer} hash
+ */
+exports.hashForEthSign = function (message) {
+  var prefix = exports.toBuffer('\u0019Ethereum Signed Message:\n' + message.length.toString())
+  return exports.sha3(Buffer.concat([prefix, message]))
+}
+
+/**
+ * ECDSA sign `message` using the same algorithm as the `eth_sign` RPC method.
+ * Prefixes message with fixed header + `message.length.toString()`, then
+ * signs the Keccak256 hash of the result.
+ * @param {Buffer} message
+ * @param {Buffer} privateKey
+ * @return {Object} signature
+ */
+exports.ethSign = function (message, privateKey) {
+  var msgHash = exports.hashForEthSign(message)
+  return exports.ecsign(msgHash, privateKey)
+}
+
+/**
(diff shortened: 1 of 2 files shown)


wanderer:
i could have sworn these where already implemented somewhere. @kumavis @axic any know?

kumavis:
@wanderer nah they have been reimplemented at the client layer (e.g. metamask)
I would be happy to have access to this

axic:
We have part of this implemented as `fromRpcSig` and `toRpcSig`. I would like to move the two closer and not have that many new methods if possible.

kumavis:
Ah ok -- I second @axic's request

yusefnapora:
Sorry, I'm not clear on how `fromRpcSig` and `toRpcSig` can help with this; it looks like they just convert the string representation of the signature returned by `eth_sign` into an object with `r`, `s`, and `v` params (and make sure that `v` is in the correct range of 27-28).  In other words, they convert an existing signature from one format to another.  

The problem is that the signature generation algorithm has changed, so there's no way to produce a signature compatible with the new `eth_sign` behavior without doing the prefixing yourself and running the prefixed message through keccak-256.  

I suppose you could change `ecrecover` and `ecsign` so they would do the prefixing and hashing, but that seems like it would dramatically break the API.  It seemed safer to add new methods rather than change the existing ones.

index.js
```diff
@@ -360,6 +360,32 @@ exports.ecsign = function (msgHash, privateKey) {
...
+
+/**
+ * ECDSA sign `message` using the same algorithm as the `eth_sign` RPC method.
+ * Prefixes message with fixed header + `message.length.toString()`, then
+ * signs the Keccak256 hash of the result.
+ * @param {Buffer} message
+ * @param {Buffer} privateKey
+ * @return {Object} signature
+ */
+exports.ethSign = function (message, privateKey) {
```
axic: I think naming this `ethSign` will cause a lot of confusion, given there's `ecsign` and this is not for signing transactions. Since it is such a slim helper, maybe it is not necessary to have it here?

index.js
```diff
@@ -378,6 +404,19 @@ exports.ecrecover = function (msgHash, v, r, s) {
...
 
 /**
+ * ECDSA public key recovery from a message signed with the `eth_sign` RPC call.
+ * @param {Buffer} message
+ * @param {Number} v
+ * @param {Buffer} r
+ * @param {Buffer} s
+ * @returns {Buffer} publicKey
+ */
+exports.ethRecover = function (message, v, r, s) {
```
axic: The same applies here as to `ethSign`.

kumavis:
@yusefnapora  can we pass the sig result from your new fn to `toRpcSig`. or does not return the correct result format any more?

index.js
```diff
@@ -360,6 +360,32 @@ exports.ecsign = function (msgHash, privateKey) {
...
 
 /**
+ * Returns the keccak-256 hash of `message`, prefixed with the header used by the `eth_sign` RPC call.
+ * The output of this function can be fed into `ecsign` to produce the same signature as the `eth_sign`
+ * call for a given `message`, or fed to `ecrecover` along with a signature to recover the public key
+ * used to produce the signature.
+ * @param message
+ * @returns {Buffer} hash
+ */
+exports.hashForEthSign = function (message) {
```
axic: I think this should be named differently, given it is not for transaction signing. Perhaps something like `hashPersonalMessage` because it initially was created as the `personal.sign` RPC method.

yusefnapora:
sure, you can send the output through `toRpcSig` - the `ethSign` function calls `ecsign` under the hood, so it will return sigs in the same format as `ecsign`.  It just does the prefixing / hashing first.

axic:
> Sorry, I'm not clear on how fromRpcSig and toRpcSig can help with this;

Sorry for that, they were also originally not named `*RpcSig`, but they were really intended to be used together with `eth.sign`.

kumavis:
@yusefnapora is the `v` supposed to be `{0,1}` or `{27,28}` in the new standard?

yusefnapora:
It looks like geth is returning `{27,28}` in the current master branch: https://github.com/ethereum/go-ethereum/blob/682875adff760a29a2bb0024190883e4b4dd5d72/internal/ethapi/api.go#L304 - I honestly don't know what the standard says 😛

It also looks like the `EcRecover` api function in geth does the prefixing and hashing before doing the actual recovery.  Which is a little confusing, since `ecrecover` in this package doesn't, but I personally wouldn't want to change the behavior of `ecrecover` here since it might break code that's depending on the old behavior.

I'm fine with getting rid of the `ethSign` and `ethRecover` methods and just having the method that does the prefixing + hashing.  Renaming that to `hashPersonalMessage` sounds good too.  Then consumers can do `ecsign(hashPersonalMessage(myMessage), privateKey)` and get the right sig.

kumavis:
`ecsign(hashPersonalMessage(myMessage), privateKey)` 👏

kumavis:
so ignore `fromRpcSig` and `toRpcSig`, we'll leave them for legacy support for now

yusefnapora:
awesome, sounds good 👍  thanks for helping me sort this out

kumavis:
groovey ☮️

yusefnapora:
@axic, @kumavis does this work for you guys with the latest change? I tried rebuilding the docs, but that changed all the links to point to my fork, so I guess you guys can do that after merging.

kumavis:
@axic please approve changes if you are satisfied

axic:
Thank you @yusefnapora, it looks good!

yusefnapora:
Awesome, thanks :)

kumavis:
@axic they are not compatible as the rpc spec has changed

`fromRpcSig` and `toRpcSig` should be deprecated

flyswatter:
This PR does not actually include `ethRecover`!

kumavis:
```js
function ethRecoverPersonalMessage(message, sig) {
  const msgHash = ethUtil.hashPersonalMessage(message)
  const sigParams = ethUtil.fromRpcSig(sig)
  const sender = ethUtil.ecrecover(msgHash, sigParams.v, sigParams.r, sigParams.s)
  return sender
}
```

kumavis:
@yusefnapora where did you get that test data from for the "Hello World" hash?

yusefnapora:
@kumavis I just generated it from the node repl using an earlier javascript implementation I wrote for the mediachain project, before I opened this PR.  Which is kind of cheating, of course.  Ideally it would be generated by geth, but I haven't gone deep enough into the geth code to do so.  Since there's no RPC call exposed for just creating the personal message hash (I think), it seems like you'd need to either write a one-off go program or use something like [gore](https://github.com/motemen/gore) to call the function used internally.

quantumproducer:
@yusefnapora does this fix https://ethereum.stackexchange.com/questions/12033/sign-message-with-metamask-and-verify-with-ethereumjs-utils ?

PR #68, opened 2017-02-10
moves hex utils to ethjs-util

wanderer:


axic:
What problem is this solving? :)

wanderer:
less code to maintain. @SilentCicero can focus on those string methods, which pretty much leaves crypto functions in ethereumjs-util

PR #70, opened 2017-02-15
add recoverPersonalMessage

kumavis:
cc @flyswatter @yusefnapora @wanderer

### DIFF
--- a/index.js
+++ b/index.js
@@ -348,6 +348,20 @@ exports.hashPersonalMessage = function (message) {
 }
 
 /**
+ * Returns the address corresponding to the `personal_sign`-formatted `signature` for the `message`.
+ * @param message
+ * @param signature
+ * @returns {Buffer} hash
+ */
+
+exports.recoverPersonalMessage = function (message, signature) {
+  const msgHash = exports.hashPersonalMessage(message)
+  const sigParams = exports.fromRpcSig(signature)
+  const sender = exports.ecrecover(msgHash, sigParams.v, sigParams.r, sigParams.s)
+  return sender
+}
+
+/**
  * ECDSA public key recovery from signature
  * @param {Buffer} msgHash
  * @param {Number} v


yusefnapora:
Sounds good to me :)

flyswatter:
I would prefer having some test data for these signature methods, to ensure we're interacting with the other clients correctly.

axic:
Can we come up with some consistent naming? It is hard to change names afterwards.

I'm not saying the current name is wrong, but maybe it would make sense spending a tiny bit of time thinking about what kind of API we have.

axic:
Also this is lacking tests (see the coverage decrease).

kumavis:
yeah im trying to find good test data first

axic:
Also I wonder if this is now a standardised format or is it still just generated by a random RPC method?

flyswatter:
This format of signature is now in both geth and parity, and was first deployed to the public in [this geth PR](https://github.com/ethereum/go-ethereum/pull/2940), named `personal_sign`, so I think the name `recoverPersonalSign` is pretty specific and accurate to the goals of this method.

axic:
I mean if it is only an RPC method, why should it be in ethereumjs-util?  If it is not an RPC method, but a standardised format, and is also available via RPC, that is a different story.

kumavis:
tbh im not sure whats in and out of scope for `ethereumjs-util`

flyswatter:
Oh sorry, recovering `personal_sign` is not an RPC method, only the signing is.  Recovering is a nice utility method.

`recoverPersonalSign` belongs in `ethereumjs-utils` for similar reasons as `hashPersonalMessage` and `ecrecover` does.

axic:
I'm not sure why recovering belongs here if signing doesn't. They do the same wrapping over the the util methods:
- hash input
- sign & wrap signature or unwrap & recover signature

flyswatter:
I'm not saying signing doesn't belong here, I'm saying recovering does.  Signing might also, but is less urgent, since it's provided by existing RPCs.

axic:
> Signing might also, but is less urgent, since it's provided by existing RPCs.

Sorry I don't follow?

index.js
```diff
@@ -348,6 +348,20 @@ exports.hashPersonalMessage = function (message) {
 }
 
 /**
+ * Returns the address corresponding to the `personal_sign`-formatted `signature` for the `message`.
+ * @param message
+ * @param signature
+ * @returns {Buffer} hash
+ */
+
+exports.recoverPersonalMessage = function (message, signature) {
```
axic: Since we require the original message here, wouldn't it make sense supplying an address and making this a `verifyPersonalMessage`?

`recoverPersonalMessage` wounds a bit like a message being recovered, while we're recovering the signing address.

flyswatter:
> Sorry I don't follow?

Right now JS devs have a way of getting hashed data using `personal_sign` from an RPC, but they have no easy util way of verifying it.  That's why I call it an urgent method.  There are current RPC methods that are not easily verified.

I actually have already written the `personal_sign` logic in JS, I could PR that against `ethereumjs-util` if you think it belongs here just as well.

flyswatter:
Oh sorry, I haven't written personal_sign in JS, I'm working on that now.  I wrote the other type of sign.

axic:
@flyswatter the issue you've linked defines a `personal_recover` RPC method, I assume it is indeed present?

flyswatter:
Oh, if it's present, I was thrown off because it [is not on the wiki](https://github.com/ethereum/wiki/wiki/JSON-RPC).  It's also [not on the parity wiki](https://github.com/ethcore/parity/wiki/JSONRPC-personal-module). Where do you see it?

flyswatter:
Oh, the issue I linked.  Sorry, I'm doing too many things at once right now.

flyswatter:
Ok, so let me start over:

The best argument for implementing this *despite it being an RPC method* is that this represents *a unique RPC method*.

This RPC method's definition was changed abruptly, with little community-facing warning, and in my opinion, it should have been more gradually phased in, with a more usual type of deprecation warning, replacement, and eventual removal.

The reason the MetaMask team is now promoting these methods in `ethereumjs-util` is because we want to provide the gradual transition that geth and parity did not to our users.  This requires having access to both methods at once.

If you think that's a provider issue, I understand that perspective, that's fine, we can just put these methods somewhere else.

kumavis:
geth implements `personal_ecRecover`

index.js
wanderer: pinging @kumavis

wanderer: oh wait on vac, :P will wait till ur back

kumavis:
relevant https://www.npmjs.com/package/eth-sig-util

wanderer:
https://github.com/flyswatter/eth-sig-util

holgerd77:
Outdated, will close.

PR #71, opened 2017-02-20
Use `addHexPrefix` in `privateToPublic`

fanatid:
With this improvement `privateToAddress` can convert `privateKey` in HEX to address.


holgerd77:
This needs some context what problem is tackled here.

fanatid:
Unfortunately this was 1.5y ago and now I do not remember details, I think can be closed :)

PR #84, opened 2017-07-17
Make sure we have a sig.recovery different than 'undefined'

imcu:


wanderer:
@imcu  could you add a test?

index.js
```diff
@@ -327,6 +327,10 @@ exports.importPublic = function (publicKey) {
 exports.ecsign = function (msgHash, privateKey) {
   var sig = secp256k1.sign(msgHash, privateKey)
 
+  if (sig.recovery === undefined) {
```
axic: In what case can this be `undefined`?

I'd count that as a failure and as such I would throw an error.

fanatid: `recovery` is always defined

axic: @imcu can you give a test case where you've received `undefined`?

imcu:
This was happening on my machine because of the way my Webpack gathered modules together. 
Please close. No longer relevant.

index.js
imcu: I followed what was happening and it's not the library's fault. 
My webpack was making sekp256k1 to load elliptic@3.0.3 instead of the elliptic@6.2.3 specified in the dependency.
So this is not a bug of the library. Please close the PR.

PR #88, opened 2017-08-07
es6 round 1 - remove all vars and anon functions

Silur:
nothin special here

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "5"
   - "6"
   - "7"
 env:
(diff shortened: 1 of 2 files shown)


wanderer:
looks good to me. @kumavis do you think this would break anything downstream for ya?

kumavis:
Js-ipfs wants node v4 support (not sure if we currently have that)

I love es7, but here it's providing minimal benefit and yet reduces platform compatibility, so I'm leaning against merging this atm.

I wish there was an elegant way to transpile and publish 

Thanks for the PR, let me think about it

kumavis:
Talked to ipfs, they have abandoned node v4 support so we're good to merge on this

wanderer:
after discussion we decided to merge

Issue #89, opened 2017-08-08
Inconsistent between `isValidAddress` and `defineProperties`


tranvictor:
In `isValidAddress`:
```
exports.isValidAddress = function (address) {
  return /^0x[0-9a-fA-F]{40}$/i.test(address)
}
```
We accept a string starting with `0X` because the regex is with `i` case insensitive flag. However `0X` is invalid hex string to `defineProperties`.

I do not think we should consider `0X` a valid prefix of a hex string.

Proposal: Just remove `i` flag in the regex. I don't see any good reasons to have it here.

axic:
I think you are right. It must have been a mistake, since the regex contains both upper/lowercase.

Could you open a pull request fixing it?

tranvictor:
@axic sure, i will propose a PR soon.

PR #90, opened 2017-08-09
do not allow 0X in address

tranvictor:


tranvictor:
refer to #89

axic:
@wanderer @kumavis @fanatid any comments?

kumavis:
just clarifying - with this PR: `0xabcd...` is ok but `0Xabcd` is not, correct?

tranvictor:
@kumavis it is correct and it is also consistent to `defineProperties` that way.

axic:
@kumavis good to go?

Issue #111, opened 2018-01-08
Throws an error while trying to minify with create-react-app


nachomazzara:
If you have `ethereumjs-util` as dependency for you webApp using `create-react-app`, once you want to run the build script: 
`npm run build` throws 

```Failed to minify the code from this file: ./node_modules/ethereumjs-util/index.js:142```

Is it possible to transpile the code with `Babel`, create a `dist` folder where we can get the transpiled code and set dist as the main place in `package.json`?

Thanks!

holgerd77:
What do you mean with "Set ``dist`` as the main place in ``package.json``?

fanatid:
@nachomazzara it is possible, but I do not think that this will be done. IIRC ethereumjs supports only node > 6.
So, you should setup browserify/webpack config for transpiling ethereumjs-util in your project.

nachomazzara:
Sorry for the delay.

@holgerd77 I mean, instead of using the index.js at the root of the repo. Use an `index.js` transpiled inside a `/dist` folder like almost every npm module have

@fanatid setting `Browserify/Webpack` is an option but if you are working with [create react app](https://github.com/facebookincubator/create-react-app) you will need to eject it and change Webpack configuration which breaks the _plug and play_

If you don't want to have the main branch with the transpiled version, We could create a #support-create-react-app branch for this purpose :)

Thanks guys

admazzola:
I get an error trying to minify with uglifer -- is this a coincidence ?

`ERROR in application-85c2f5c841fd6db557e9.js from UglifyJs
Unexpected token: name (first) [./node_modules/ethereumjs-util/index.js:142,0][application-85c2f5c841fd6db557e9.js:29540,6]

ERROR in admin-controls-interface-d1f6a968ffa463219e33.js from UglifyJs
Unexpected token: name (first) [./node_modules/ethereumjs-util/index.js:142,0][admin-controls-interface-d1f6a968ffa463219e33.js:29540,6]

ERROR in good-contract-interface-3cfb8b306b0e6eefe793.js from UglifyJs
Unexpected token: name (first) [./node_modules/ethereumjs-util/index.js:142,0][good-contract-interface-3cfb8b306b0e6eefe793.js:29540,6]

ERROR in ethereum-account-manager-79a41ffe50545abe549e.js from UglifyJs
[...]

Issue #92, opened 2017-08-26
Can't import ethrereumjs-util in React Native


amaurymartiny:
If I create a fresh React Native app with `create-react-native-app`, and only add the line:

```js
import utils from 'ethereumjs-util';
```

I get an error:

![screen shot 2017-08-26 at 5 34 52 pm](https://user-images.githubusercontent.com/1293565/29742763-eae154b4-8a84-11e7-8509-dc3c4761a75c.png)

Edit: After installing the packages manually `yarn add assert`, and then `yarn add stream` and `yarn add buffer`, I get the following error:

![screen shot 2017-08-26 at 5 33 39 pm](https://user-images.githubusercontent.com/1293565/29742753-bfcf9876-8a84-11e7-92b4-3fd6d770affb.png)

tharzeez:
Any solution for this. I tried with rn-nodeify 
[https://www.npmjs.com/package/rn-nodeify](url)

but still I am getting error buffer not defined .. Tried npm install buffer still no hope

kouohhashi:
I have the same issue.
Have you found a solution?

tharzeez:
I followed these steps and in step 9 : 
[https://gist.github.com/dougbacelar/29e60920d8fa1982535247563eb63766](url)
instead of `npm install --save web3`
i installed web3 1.0.0 `npm install --save web3@1.0.0-beta.24`
all those util helper class are available under a single module web3 1.0.0 so yea i can use those util files from the above package.

I hope it helps.

axic:
Fixed by #114.

Ouabi54:
I still not understand how to fix this issue

PR #114, opened 2018-01-29
Added babel-cli to distribute ES5 version on npm

holgerd77:
Adresses #111 and #92


nickysemenza:
I'd love to see this merged 😄

holgerd77:
@nickysemenza This is now in review, hope to have this merged soon.

test/defineFields.js
```diff
@@ -1,5 +1,5 @@
 var assert = require('assert')
-var ethUtil = require('../index.js')
+var ethUtil = require('../dist/index.js')
```
Silur: maybe tests can be part of the babel space too for some may write these in ES6?

holgerd77: Ah, yes, we can think about that, but let's leave this for now.

.babelrc
```diff
@@ -0,0 +1,7 @@
+{
+  "presets": [
+    [
+      "env"
```
fanatid: If publisher will build with latest node this will produce not ES5 output, is not it?

holgerd77: Could you be a bit more specific, who is the "publisher", the one who publishes this on npm? And what "latest" node do you mean, which version? What changed in that version, do you have some link on that?

fanatid: Sorry, one missed word destroyed my message, I wanted write: `If publisher will build with latest node version...`

I checked https://www.npmjs.com/package/babel-preset-env
> Without any configuration options, babel-preset-env behaves exactly the same as babel-preset-latest (or babel-preset-es2015, babel-preset-es2016, and babel-preset-es2017 together).

Did not knew about this. Sorry for false alarm.

holgerd77: No problem, I am not totally the transpilation/babel expert, so better to double check than one check to few. Sorry also for being a bit unconsistent in what Node version we actually support. I think we still have to fine-tune this a bit. I also didn't know how many people still need some ``ES5`` compatible distribution and for what reasons.

So I think it probably makes sense for the various libraries to -yes- drop Node 4 support for the source version to be able to implement with newer and more elegant techniques, but also provide a transpiled distribution version with lower requirements.

Probably also has to be decided/evolved on a case-by-case (library-by-library) basis, depending on the needs of the users, the scope of the library and the compromises which have to be done in respect to security, speed and build/library size.

Actually this topic gets more and more complex for me the more I read on it. Will try to keep pace! 😄

gre:
Is there any chance this will get backport to the `4.x` release? many library still depend on the 4.x , like `ethereumjs-abi` and it will take time for everyone to update to it.

holgerd77:
@gre Hmm, not very probable, sorry, there is so much to do on the various libs, don't think anyone will find the time. Is ``ethereumjs-abi`` your main concern? I would rather look into updating this instead then.

gre:
ok thanks. I've sent a PR for -abi https://github.com/ethereumjs/ethereumjs-abi/pull/59 , i'm going to see which other dependency I have still uses the 4.x.
cheers

Issue #110, opened 2018-01-05
isPrecompiled() needs to be updated to reflect byzantium


axic:
See https://github.com/ethereumjs/ethereumjs-util/blob/master/index.js#L519, which limits it to address 0 to 5.

(This feature is used by the -vm and -accounts.)

axic:
Update: 
- apparently it is not used by `accounts` anymore for a while: https://github.com/ethereumjs/ethereumjs-account/pull/2
- the VM does its check internally

This feature could be removed in the next breaking change, but for now it still needs to be updated.

holgerd77:
Yes, removed this from the VM when working on Byzantium. Found this too implicit with too few benefits.

PR #115, opened 2018-02-02
Updated isPrecomile according to Byzantium precompile-used address space

holgerd77:
Addresses #110

### DIFF
--- a/index.js
+++ b/index.js
@@ -510,13 +510,13 @@ exports.generateAddress = function (from, nonce) {
 }
 
 /**
- * Returns true if the supplied address belongs to a precompiled account
+ * Returns true if the supplied address belongs to a precompiled account (Byzantium)
  * @param {Buffer|String} address
  * @return {Boolean}
  */
 exports.isPrecompiled = function (address) {
   const a = exports.unpad(address)
-  return a.length === 1 && a[0] > 0 && a[0] < 5
+  return a.length === 1 && a[0] >= 1 && a[0] <= 8
 }
 
 /**
--- a/test/index.js
+++ b/test/index.js
@@ -411,11 +411,17 @@ describe('isPrecompiled', function () {
     assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000002'), true)
     assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000003'), true)
     assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000004'), true)
+    assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000005'), true)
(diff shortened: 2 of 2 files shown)


holgerd77:
@Silur Hi, thanks for the quick ES5 PR review, can you also do a short review of this (nothing exciting)? Would like to have this included before we do a release.

index.js
```diff
@@ -510,13 +510,13 @@ exports.generateAddress = function (from, nonce) {
 }
 
 /**
- * Returns true if the supplied address belongs to a precompiled account
+ * Returns true if the supplied address belongs to a precompiled account (Byzantium)
  * @param {Buffer|String} address
  * @return {Boolean}
  */
 exports.isPrecompiled = function (address) {
   const a = exports.unpad(address)
```
fanatid: Add `isValidAddress`?

test/index.js
```diff
@@ -411,11 +411,17 @@ describe('isPrecompiled', function () {
     assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000002'), true)
     assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000003'), true)
     assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000004'), true)
+    assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000005'), true)
+    assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000006'), true)
+    assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000007'), true)
+    assert.equal(ethUtils.isPrecompiled('0000000000000000000000000000000000000008'), true)
+    assert.equal(ethUtils.isPrecompiled(Buffer.from('0000000000000000000000000000000000000001', 'hex')), true)
```
fanatid:
```js
for (let i = 1; i < 9; ++i) {
  assert.equal(ethUtils.isPrecompiled(new BN(i).toString(10, 40)), true)
}
```

index.js
holgerd77: Hmm, normally your are right. I wouldn't want this here, since the ``IsPrecompiled`` method is removed anyhow (citing @axic) on next major release. Adding ``isValidAddress`` would change the behavior here and would cause extra/unnecessary problems for people already have this in use.

test/index.js
holgerd77: This is definitely more elegant, thanks for the suggestion. Unfortunately it kills the browser tests somehow due to Buffer usage, just tested it, and I didn't want to go deeper in debugging here, think this would be not worth it in this case. Is it ok to leave it as is?

Also, can you approve, If you are ok with both explanations?

fanatid: Yea, this is ok let it as is.

holgerd77: Cool, thanks.

PR #127, opened 2018-04-26
New release v6.0.0 (abandoned)

holgerd77:
Have chosen a major version bump on this to incentivise people to do the naming updates and so to minimize risk of breaking on the next major release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [6.0.0] - 2018-04-27
+- Rename all ``sha3`` hash related constants and functions to ``keccak``, see
+  [this](https://github.com/ethereum/EIPs/issues/59) EIP discussion for context
+  (tl;dr: Ethereum uses a slightly different hash algorithm then in the official
+  ``SHA-3`` standard)
+- Renamed constants:
+  - ``SHA3_NULL_S`` -> ``KECCAK256_NULL_S``
+  - ``SHA3_NULL`` -> ``KECCAK256_NULL``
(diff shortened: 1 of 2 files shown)


axic:
Probably we should add an alias of `keccak256(x)` to `keccak(x, 256)`

holgerd77:
Yes, that makes sense. I can directly create a PR on this, or are you already doing that?

axic:
I'm not doing, please feel free to create a PR for it.

holgerd77:
Ok, have done, maybe have a short look: https://github.com/ethereumjs/ethereumjs-util/pull/128

axic:
Also need to remove the `sha3` alias before the release.

axic:
The point of a breaking release is to remove backwards compatible changes. We had `sha3` as an alias for a couple of releases now while having `keccak` in the API.

There is no need to make a breaking release if `sha3` stays.

holgerd77:
No, there was not a single release with the ``keccak`` aliases yet. My reasoning here that people tend to "overlook" minor releases. So they wouldn't stumble on the deprecation notes from the release notes (cause they don't read them). If we make a major release here chances are higher that people have this in place once v7.0.0 comes out with removed aliases.

I have no strong opinion here though - would regard the above as a good way to go - but we can also
a) directly make a major releases with aliases removed
b) do a minor release still including the aliases

Don't mind.

axic:
If you keep `sha3` in the next release then it is not breaking. I think the best is to make a regular non-breaking release to introduce `keccak` and `keccak256`.

axic:
Also note the point of semantic versioning is that people won't be hit with a breaking build. They just have to manually upgrade to 6.0.0 and see what changed.

holgerd77:
Ok, then I would prepare a ``v5.2.0`` release instead just introducing the new ``keccak`` versions and keep the aliases and deprecation note.

Offtopic question: can we/you/someone else do an ``ethereumjs-abi`` release today?

Issue #124, opened 2018-03-02
Remove setLength and sha3 in next breaking release


axic:

Anaphase:
Yeah, the docs say `sha3` is obsolete and 121b4d9 shows that `keccak` should be used in it's place. However, this change hasn't been published to npm yet ¯\\\_(ツ)_/¯. The docs should reflect what's currently published; might be time for 6.0.0?

holgerd77:
@Anaphase Thanks for the reminder, have prepared a PR for a release on this: https://github.com/ethereumjs/ethereumjs-util/pull/127

holgerd77:
Note on this: sha3 references has been removed in the ``v6.0.0`` release. A bit unfortunate, didn't remove ``setLength`` (see [here](https://github.com/ethereumjs/ethereumjs-util/blob/master/index.js#L106) in the code), this would still have to be done in the next major release.

I would interpret the change to be done as: ``setLength()`` has been replaced with ``setLengthLeft()`` at some point in time, the ``setLength()`` alias should be removed, please correct me if I am wrong.

axic:
> I would interpret the change to be done as: setLength() has been replaced with setLengthLeft() at some point in time, the setLength() alias should be removed, please correct me if I am wrong.

Correct.

axic:
`sha3` was removed in #154, but `setLength` is still in.

holgerd77:
`setLength` has been removed in #244 and published along the [v7.0.0](https://github.com/ethereumjs/ethereumjs-util/releases/tag/v7.0.0) release, will close.

PR #38, opened 2016-04-30
Split off secp256k1 into its own file [WIP]

axic:
Fixes #35.

Dont merge yet - not fully happy with it.

PR #170, opened 2019-01-03
Migrate to typescript

s1na:
This PR migrates the library to typescript. Part of the diff is only formatting, and due to running `prettier` on the codebase.

The PR needs further testing and is work in progress.


src/index.ts
```diff
@@ -560,27 +577,36 @@ exports.isPrecompiled = function (address) {
...
 
-  return exports.isHexPrefixed(str) ? str : '0x' + str
+  return ethjsUtil.isHexPrefixed(str) ? str : '0x' + str
 }
 
 /**
  * Validate ECDSA signature
  * @method isValidSignature
- * @param {Buffer} v
+ * @param {Number} v
```
axic: Is this a breaking change introduced here or just fixing the docs?

src/index.ts
```diff
@@ -591,14 +617,14 @@ exports.isValidSignature = function (v, r, s, homestead, chainId) {
     return false
   }
 
-  r = new BN(r)
-  s = new BN(s)
+  const rBN: BN = new BN(r)
+  const sBN: BN = new BN(s)
```
axic: In `generateAddress2` shadowing variables is not a problem. Any reason to make a distinction here?

src/index.ts
```diff
@@ -418,24 +441,24 @@ exports.toRpcSig = function (v, r, s, chainId) {
  * @param {String} sig
  * @return {Object}
  */
-exports.fromRpcSig = function (sig) {
-  sig = exports.toBuffer(sig)
+export const fromRpcSig = function(sig: string) {
+  const buf: Buffer = toBuffer(sig)
```
axic: Same question here.

src/index.ts
s1na: Fixing the docs. `v` seems to be number in all other functions (and tests).

src/index.ts
s1na: I couldn't find how to shadow a variable with a different type without the typescript compiler complaining. In `generateAddress2` the type stays the same.

axic: If the type stays the same then there's no point to have that expression in the first place :)

s1na:
@krzkaczor This library uses [documentation](https://www.npmjs.com/package/documentation) to generate docs. However `documentation` doesn't support typescript yet from what I've gathered (https://github.com/documentationjs/documentation/issues/282).

Are you aware how the typescript community usually generate docs? Is [typedoc](https://github.com/TypeStrong/typedoc) a good alternative? I'm trying to use it in combination with [typedoc-plugin-markdown](https://github.com/tgreyuk/typedoc-plugin-markdown), which for example generates such an [output](https://github.com/tgreyuk/typedoc-plugin-markdown/blob/master/examples/out/README.md).

holgerd77:
@s1na Just stumbled above exactly the same on the ``common`` library transition. 😄 😄 

Would also have chosen ``TypeDoc`` for the docs, have no deep insight though yet, maybe @krzkaczor can confirm if this is a solid choice.

krzkaczor:
@holgerd77 @s1na getting to know docs generating tools for TS is on my todo list 😆 I heard good things about `TypeDoc`, so I guess it is a right choice!

s1na:
@holgerd77 I'm trying to specify the type of functions which accept parameters or return values with multiple possible types, and I'm wondering what's the best strategy here.

- It's possible to just use unions (`Buffer | string`) and keep the API as it is.
- Break up functions to make it clear what will be the return type, given a parameter type. E.g. `baToJSON` accepts `Buffer | Array` and returns `Array | string | null`, which can be broken into `baToJSON(Buffer): string` and `baToJSON(Array<?>): Array<?>` (I'm not sure what specific array type it should support).
- My personal preferred method would be to accept only a main type and not try to support every possible type, and then provide functions to easily convert between types. E.g. many of the functions accept `any` parameter and call `toBuffer` internally. They can all be modified to accept only `Buffer`, and have users of lib to make sure they convert to Buffer. This simplifies logic, and removes type dependent inconsistencies. This however entails introduces breaking API change.

I've seen a few such inconsistencies so far:

- `setLengthLeft` accepts `Buffer | Array` and supposedly returns `Buffer | Array`, but I think it never actually returns `Array`.
- `unpad` removes `0x` from strings if they have it, but doesn't prepend it back after unpadding (as `Buffer | Array` don't need this).
- `generateAddress2` docs said it accepts only `Buffer`, but it was calling `toBuffer` and the tests were passing `string`.

src/index.ts
```diff
@@ -1,179 +1,221 @@
...
+
+export type Bufferable =
+  | Buffer
+  | string
+  | number
+  | null
+  | undefined
+  | BN
+  | BufferableArray
+  | Arrayable
```
holgerd77: Types and interfaces in the `RLP` implementation have gotten their own [types.ts](https://github.com/ethereumjs/rlp/blob/master/src/types.ts) file. I'm not deep enough into `TypeScript` yet to judge what is better / best practice. So just asking: what is your reasoning here to do it this way?

src/index.ts
```diff
@@ -1,179 +1,221 @@
...
  * @var {Object}
  */
-exports.secp256k1 = secp256k1
+export { secp256k1 }
 
 /**
  * Returns a buffer filled with 0s
  * @method zeros
  * @param {Number} bytes  the number of bytes the buffer should be
  * @return {Buffer}
```
holgerd77: Just as a note, maybe you just haven't gotten to that yet: types in doc comments have to be removed, see [TSDoc](https://github.com/Microsoft/tsdoc) standard.

holgerd77:
To your multi-type-specification question:

I am also very much in favor of your preferred approach and I think opinions stated in the past from e.g. @axic or @whymarrh also stated very much in that direction, so I think we should go on here and get the API (more) right and introduce breaking changes where necessary.

I think we should nevertheless do/discuss this on a case-by-case (method-by-method) basis, since the nature and usage setting of the different functions differ very much. So can you post here which method(s) you want to change to what new method signature and/or split-up functionality here so that we can get to a short adjustment/assurance before this is actually changed?

Maybe do one post with edits and a curated list of API changes as well as checkboxes if pushed? Then we don't loose oversight. Or otherwise around: I will directly start on the next comment with some suggestions, then we can take this as a starting point.

holgerd77:
## Breaking API Changes

### setLengthLeft
Also just did some organization-wide [search](https://github.com/search?q=org%3Aethereumjs+setLengthLeft&type=Code), this is in (EthereumJS) practice also mostly used on `Buffer` anyhow (doesn't mean there are other usages out there).

Would suggest that we drop the `Bufferable` (love the naming 😄) support here and just return `Buffer` here (respectively change the docs / return type to reality).

The `setLength` method can only be removed on this occasion (see [here](https://github.com/ethereumjs/ethereumjs-util/issues/124) since we will be doing a breaking release anyhow. We just have to remember to also state this in the release notes.

TODO: Other methods, please read for latest version on GitHub and not via mail

holgerd77:
Also put @vpulim in cc here for comments.

holgerd77:
Ok, I think I am through with suggestions on API changes, see above. I would especially love to hear some explicit statement from @axic if time permits, but also e.g. @krzkaczor being the TypeScript type expert and everyone else. This should be broadly discussed and confirmed/refined.

holgerd77:
@s1na What do you think?

src/index.ts
s1na: Yeah, I still have to do that. I kept them around for now to help me with annotating the types :)

s1na:
@holgerd77 Great! Thanks for the detailed suggestions. I think they make sense. Just a few comments:

- If I'm not mistaken, because typescript types are statically checked and not during runtime, **HexString** can only be implemented as a class, which means users must convert to/from for function invocations. Just writing this to be taken into consideration, I don't have a preference here.
- For **unpad** we can also keep the same name but have three signatures.
- If **isPrecompiled** is HF-dependent, maybe it can be moved to `ethereumjs-common`? (that is, if it has use-cases and we don't want to drop it).
- There are still cases where they accept only `Buffer`, but still call `toBuffer`, e.g. **bufferToInt**. Should we keep them as they are, or rather throw an error if they receive an unexpected type? I believe this helps with discovering bugs.

src/index.ts
holgerd77: Have no strong feeling about it, if you want just leave as is. Some reorg likely also makes sense on this library - see also e.g. https://github.com/ethereumjs/ethereumjs-util/pull/38. We should eventually do this in a second round though, think there is enough to concentrate on here.

holgerd77:
@s1na Numbered your stuff for easier reference.

1. Ah, yes, that doesn't make sense to require users to convert. Then we should at least pass input params to a dedicated check method and throw if no hex-prefix provided, and always return prefixed hex strings. Does this make more sense? Or any other options (anyone?)?

2. Ok, that's more convenient and makes more sense.

3. Yes, might make sense - at least the right place there - this would need some additional structure in the JSON files. Not sure if it's worth the effort, but we can keep this in mind (or directly open an issue).

4. I think these conversions should be removed. We should very much have a look that we keep track in the post above though, so that we document all that stuff and list all breaking API changes.

holgerd77:
@s1na Before you start nevertheless let's wait 1-2 days more for some more opinions on this.

axic:
> Ok, I think I am through with suggestions on API changes, see above. I would especially love to hear some explicit statement from @axic if time permits

I will need to properly read it, but I would prefer if it would be possible to make the agreed breaking changes in Javascript and make a last release with that, before merging the Typescript rewrite. This decouples the two changes and hopefully reduces the potential bugs/issues.

Does that makes sense?

holgerd77:
No, not really in my opinion. `TypeScript` is a perfect fit here to actually REDUCE the level of potential bugs since one is guided and safe-guarded by the type system. So would say we should stick with the combination of changes.

holgerd77:
Other opinions on that?

s1na:
@holgerd77 would it maybe make the review process easier if for now set the ambiguous types to `any`, and do the breaking changes in a separate PR after this one?

holgerd77:
@s1na Yes, probably.

holgerd77:
This also lacks displaying the main code changes in a renaming context, which makes it too hard to review and breaks file history. Can you see if you can do this two-step process here as well?

Also: I know this is a bit annoying and will probably cause an extra half-an-hour of work. But commit history got really messy here with all the changes and re-changes. Can you re-structure this into some clean commits and then do a force-push on this?

I think in this case it would be really worth the effort.

s1na:
Yeah it had gotten messy, sorry about that :sweat_smile: I rewrote the commit history, hopefully it's more clean now.

This time I tried your suggestion and first moved `index.js` to `src/index.js` in one commit, then renamed it to `src/index.ts` in a second commit, and applied to modifications in a third commit. But github is still showing it as a deleted and newly added file.

However, this time I applied all of the changes to `index` in one commit. You can use its [diff](https://github.com/ethereumjs/ethereumjs-util/pull/170/commits/fa45a4f4db203625606855510a5c904abd8f9258) to review the actual transition of source code.

holgerd77:
@s1na All global constants from the `index.ts` file are now also added to the docs, e.g. [Buffer](https://github.com/ethereumjs/ethereumjs-util/blob/852222b8801ab052771d1d9a1020aa8e54bf3a0a/docs/README.md#buffer).

Can you make sure (hopefully this is not a TSDoc bug?) that this is not the case and only the exported constants and functions are included?

package.json
```diff
@@ -3,19 +3,26 @@
...
     "test": "npm run lint && npm run test:node && npm run test:browser",
-    "test:browser": "npm run build:dist && karma start karma.conf.js",
-    "test:node": "npm run build:dist && istanbul test mocha -- --reporter spec",
-    "build:dist": "babel index.js --source-root ./ -d ./dist",
-    "build:docs": "documentation build ./index.js --github --sort-order='alpha' -f md > ./docs/index.md"
+    "test:browser": "npm run build && karma start karma.conf.js",
+    "test:node": "npm run build && istanbul test mocha -- --reporter spec",
+    "tsc": "tsc --noEmit",
+    "tslint": "tslint -p ./tsconfig.json -e node_modules/**/* -e **/node_modules/**/* -e dist/**/* **/*.ts",
+    "tslint:fix": "tslint --fix --format stylish -p ./tsconfig.json -e node_modules/**/* -e **/node_modules/**/* -e dist/**/* **/*.ts"
```
holgerd77: Can we already use the new unified script command here and for the different commands and update the associated `ethereumjs-config-*` dev dependencies?

package.json
```diff
@@ -3,19 +3,26 @@
...
+  "types": "./dist/index.d.ts",
   "files": [
     "dist"
   ],
   "scripts": {
-    "coverage": "npm run build:dist && istanbul cover _mocha",
+    "build": "tsc -p ./tsconfig.prod.json",
+    "prepublishOnly": "npm run test && npm run build",
+    "coverage": "npm run build && istanbul cover _mocha",
     "coveralls": "npm run coverage && coveralls <coverage/lcov.info",
```
holgerd77: Can you also use [ethereumjs-config-nyc](https://github.com/ethereumjs/ethereumjs-config/tree/master/packages/ethereumjs-config-nyc)?

s1na: This library uses mocha for testing and tests against the `dist/` files, and further runs karma. I tried to make it work with the nyc scripts with no luck so far. I think it would require larger changes, and probably would need to be tested against `src/` files for coverage with `ethereumjs-config-nyc` configs to work. Should I attempt the changes in this PR?

s1na:
Regarding documentation, added `--excludeNotExported` which fixes the issue you mentioned. Although I noticed there's another issue. The libraries we're re-exporting are not documented (`BN`, `rlp` and `secp256k1`). However I haven't yet found a work-around for this, there are a few issues on typedoc which are related to this, e.g. https://github.com/TypeStrong/typedoc/issues/712.

package.json
holgerd77: Ah no, then leave for now. Sorry, didn't have a close-enough look.

holgerd77:
Hmm, seems we have to deal with this somehow for now since these are open issues on `TypeDoc`. I very much would like to prevent to manually post-edit the docs, this always went unreliable from past experiences.

Can we live with some addition to the [README API section](https://github.com/ethereumjs/ethereumjs-util#api), something like:

---

Re-exports of other libraries:

- `BN` ([BN.js](https://github.com/indutny/bn.js/))
- ...

[Link to internal docs]

---

Together with the explicit library linking this might even be some sort of improvement.

If you're ok with it can you add? Thanks!

s1na:
Yeah, that's probably the best approach for now. Added to README.md.

holgerd77:
Hi Sina, sorry, I merged https://github.com/ethereumjs/ethereumjs-util/pull/171 without thinking about the implications for the `TypeScript` branch. Can you rebase on top of this respectively integrate the changes? Hope this is not causing too much trouble... 😕

I will directly take on this for review after the update.

src/index.ts
```diff
@@ -0,0 +1,714 @@
...
+/**
+ * Keccak-256 hash of the RLP of null
+ */
+export const KECCAK256_RLP_S: string =
+  '56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'
+
+/**
+ * Keccak-256 hash of the RLP of null
+ */
+export const KECCAK256_RLP: Buffer = Buffer.from(KECCAK256_RLP_S, 'hex')
```
holgerd77: Checked for equivalency of the constants, ok.

src/index.ts
```diff
@@ -0,0 +1,714 @@
...
+/**
+ * [`rlp`](https://github.com/ethereumjs/rlp)
+ */
+export { rlp }
+
+/**
+ * [`secp256k1`](https://github.com/cryptocoinjs/secp256k1-node/)
+ */
+export { secp256k1 }
+
```
holgerd77: Did some short manual tests on re-export availability, ok.

src/index.ts
```diff
@@ -0,0 +1,714 @@
...
+  if (Buffer.isBuffer(ba)) {
+    return `0x${ba.toString('hex')}`
+  } else if (ba instanceof Array) {
+    const array = []
+    for (let i = 0; i < ba.length; i++) {
+      array.push(exports.baToJSON(ba[i]))
+    }
+    return array
+  }
+}
```
holgerd77: Methods equivalent, ok.

src/index.ts
```diff
@@ -0,0 +1,714 @@
...
+      const keys = Object.keys(data)
+      fields.forEach((field: any) => {
+        if (keys.indexOf(field.name) !== -1) self[field.name] = data[field.name]
+        if (keys.indexOf(field.alias) !== -1) self[field.alias] = data[field.alias]
+      })
+    } else {
+      throw new Error('invalid data')
+    }
+  }
+}
```
holgerd77: Did a line-by-line comparison of `defineProperties`, should be ok.

holgerd77:
Ok, have merged. Especially love this [new documentation style](https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/README.md), think particularly for this library `TypeScript` is a huge huge win also with the types available when coding and using the functionality. 🎊

@s1na I think a `v6.1.0` release should do it here, what do you think?

s1na:
Great! Yeah I think `v6.1.0` should be okay, as no breaking change was introduced here.

PR #171, opened 2019-01-04
Fix isValidSignature method

danjm:
This PR makes two changes.

This first is minor: the docs for `isValidSignature` say that `v` should be a `Buffer`, but the code assumes it is a `number`.  This PR updates the doc, as changing the code would be a breaking change.

The second is a more significant bug. Prior to this PR, `isValidSignature` will return false if passed an s-value greater than secp256k1n/2 AND if `homestead === false`. However, the EIP that introduced this https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.md was the primary homestead EIP. The code seems much more correct if we instead check that homestead is true. This is consistent with what is currently done in `ethereumjs-tx` (for example). This PR updates that line of code accordingly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -598,7 +598,7 @@ exports.isValidSignature = function (v, r, s, homestead, chainId) {
     return false
   }
 
-  if ((homestead === false) && (new BN(s).cmp(SECP256K1_N_DIV_2) === 1)) {
+  if (homestead && (new BN(s).cmp(SECP256K1_N_DIV_2) === 1)) {
     return false
   }
 
--- a/index.js
+++ b/index.js
@@ -571,7 +571,7 @@ exports.addHexPrefix = function (str) {
 /**
  * Validate ECDSA signature
  * @method isValidSignature
- * @param {Buffer} v
+ * @param {Number} v
  * @param {Buffer} r
  * @param {Buffer} s
  * @param {Boolean} [homestead=true]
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -598,7 +598,7 @@ exports.isValidSignature = function (v, r, s, homestead, chainId) {
     return false
   }
 
-  if ((homestead === false) && (new BN(s).cmp(SECP256K1_N_DIV_2) === 1)) {
+  if (homestead && (new BN(s).cmp(SECP256K1_N_DIV_2) === 1)) {
```
holgerd77: Staring at this for some time now. Actually in the docs it is noted that ``homestead`` is set to ``true`` by default, however this is enforced nowhere in the code. My theory that this went unnoticed is the following: the ``chainId`` introduction is pretty new, just lately with the [v6.0.0](https://github.com/ethereumjs/ethereumjs-util/releases/tag/v6.0.0) release. I would assume that people before just have used this without setting ``homestead`` explicitly, assuming that the docs comment is correct. By having this check inverted this was then actually acting in a homestead way by a two-false-boolean double flip.

I would therefore suggest that we do this double change and add the ``homestead`` default setting along with this, I would assume this should play out in combination. 😄

Does this make sense? If you go along just update the PR or otherwise note where you disagree.

danjm: I agree with your above assessment and updated the code accordingly 777241f

I also renamed the `homestead` param to `homesteadOrLater` as the validation rule is meant to apply to homestead and all later forks.

PR #174, opened 2019-01-28
changed '0x0' buffer conversion from <00> to <>

zmitton:
The response coming from both Geth and Parity for an empty byte array is `"0x0"`. The function `util.toBuffer("0x0")` should therefore return `<buffer >`, instead of `<buffer 00>` to match rpc response behavior.

Reproduction step:
==============
`curl -X POST --data '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0x100",false],"id":1}' -H "Content-type:application/json" https://mainnet.infura.io | jq`

The returned block includes the line: `"gasUsed": "0x0"` but an empty byte array is used in the block encoding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,5 +65,5 @@ node_modules
 package-lock.json
 
 # Build folder
-dist
+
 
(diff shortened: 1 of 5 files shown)


s1na:
This might be more relevant in [ethrpc](https://github.com/ethereumjs/ethrpc). `ethereumjs-util` is used internally in most of ethereumjs projects and changing the encoding here could lead to changes in behaviour. Unless there's good reason to also use `<buffer >` internally?

zmitton:
that's probably true, best to leave it alone

PR #179, opened 2019-02-19
Mv constants to own file

s1na:
Follow up on #178. It moves constants to their own file (to prevent accidental changes), and re-exports them from `index.ts`. Tested by manually injecting into `ethereumjs-account`.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,0 +1,50 @@
+import BN = require('bn.js')
+
+/**
+ * The max integer that this VM can handle
+ */
+export const MAX_INTEGER: BN = new BN(
+  'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff',
+  16,
+)
+
+/**
+ * 2^256
+ */
+export const TWO_POW256: BN = new BN(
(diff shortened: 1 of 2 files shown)


holgerd77:
Ah, we maybe have some more general `coveralls` problem - eventually some expired token or something? - will have a look.

holgerd77:
Ah, maybe do a switch to `nyc` on first run and see if this helps?

s1na:
Seems like other people are experiencing the same [issue](https://github.com/lemurheavy/coveralls-public/issues/632) (scroll down to the end)

holgerd77:
@s1na: Ok, will do admin-merges for now if that's the only task failing so that we can move forward till this is resolved.

holgerd77:
If it was a deliberate decision to use the singular `constant` version for the file naming I would be ok with it. I would find `constants.js` a bit more intuitive. But again: if this is following some own preference and/or naming convention I am not aware of this will also be ok.

s1na:
It was just out of habit :) I agree the plural is more intuitive usually, updated.

holgerd77:
Will do admin-merge due to current brokenness of the coveralls service.

PR #180, opened 2019-02-19
Fix broken links in docs/README.md

robiiinos:
According to README.md the link should lead to the `master` blob. But in the /docs/ folder, the README.md links lead to an old and unreachable URL (404; in blob `dd56e02` in of `master`).

This PR fix this.
Thanks for your time and the library 💯

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -71,7 +71,7 @@
 
 **● KECCAK256_NULL**: *`Buffer`* =  Buffer.from(KECCAK256_NULL_S, 'hex')
 
-*Defined in [index.ts:42](https://github.com/ethereumjs/ethereumjs-util/blob/dd56e02/src/index.ts#L42)*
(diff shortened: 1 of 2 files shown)


robiiinos:
If you plan on merging this PR, please advice me on how I could fix the failed CI test... Thanks

s1na:
Hey @robiiinos, thanks for the PR!

Those docs are auto-generated via `npm run docs:build`. Even if we fix them manually now, the next time they're generated the fix will be overwritten. It seems the links are anchored to a specific commit, I'm not sure why!

Btw, regarding the CI failure, we're experiencing a problem with coveralls since 2 days, hopefully it's resolved soon.

s1na:
Update: I think the reason the links are not working is that the commit history has been modified after the docs were generated (e.g. via `rebase`).

robiiinos:
Update : I found on we can get this done without having to build the docs at each commit; I'm closing this PR to open a new one that will fix this definitely.

Thanks for pointing the way !

PR #181, opened 2019-02-20
Fix typedoc URL generation (for /docs)

robiiinos:
This PR replace #180 

According to the https://github.com/TypeStrong/typedoc#theming documentation, we're able to specify the branch to stick with when generating docs. According to the current README.md, I specified the master branch.

I also re-build the docs to fix all of this.
Thanks @s1na for helping me out !

PS : We should be able to keep the docs generated as it is until you change the *.ts file

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -71,7 +71,7 @@
 
 **● KECCAK256_NULL**: *`Buffer`* =  Buffer.from(KECCAK256_NULL_S, 'hex')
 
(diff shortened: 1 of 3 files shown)


s1na:
That was fast!

Ah, this PR seems to also fix an issue with union types, I imagine because you have a newer version of typedoc installed and they've fixed this.

This already looks good to me. I think having the docs point to `master` instead of a specifi commit is more flexible and less likely to break. And in the worst case, if the docs are out-dated, they would point to a wrong line. I think it's a good trade-off. @holgerd77 what do you think?

PR #182, opened 2019-02-21
Structure related functions in separate source files

s1na:
This PR follows up on the discussion in #170 (and supersedes #38) and puts conceptually related functions in separate files. I totally made up the structure. Feel free to propose different files, or suggest moving specific functions from one file to another. Here's a gist:

- **index**: Only re-exports functions defined in other files + re-exports external modules (`BN`, `rlp` and `secp256k1`).
- **common**: Utility functions such as `Buffer`, byte array or string manipulation
- **pubkey**: Public-key cryptography, including managing private and public keys, and signatures
- **hash**: Hash functions
- **address**: Creating, validating and converting addresses. There is more overlap between this and `pubkey`

I haven't modified the functions themselves, except for explicitly using `ethjsUtil` in `common` instead of `exports`.

Update: I also removed the `safe-buffer` dependency.


src/address.ts
```diff
@@ -0,0 +1,107 @@
...
+  return address.slice(-20)
+}
+
+/**
+ * Returns true if the supplied address belongs to a precompiled account (Byzantium).
+ */
+export const isPrecompiled = function(address: Buffer | string): boolean {
+  const a = unpad(address)
+  return a.length === 1 && a[0] >= 1 && a[0] <= 8
+}
```
holgerd77: I think all this fits really well.

src/common.ts
```diff
@@ -0,0 +1,270 @@
...
+      const keys = Object.keys(data)
+      fields.forEach((field: any) => {
+        if (keys.indexOf(field.name) !== -1) self[field.name] = data[field.name]
+        if (keys.indexOf(field.alias) !== -1) self[field.alias] = data[field.alias]
+      })
+    } else {
+      throw new Error('invalid data')
+    }
+  }
+}
```
holgerd77: Most of this if buffer-related utility functionality - this will especially be true once we do the breaking API changes becoming more strict on implicit buffer conversions. Wonder if ``buffer.js`` would b a better fit for everything (think so?) here with the exception of ``defineProperties``.

Since we want to get rid of ``defineProperties`` on mid-term eventually anyhow, maybe it makes sense to directly isolate this on this occasion and use an extra file ``object.js`` or similar?

src/hash.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Returns the keccak-256 hash of `message`, prefixed with the header used by the `eth_sign` RPC call.
+ * The output of this function can be fed into `ecsign` to produce the same signature as the `eth_sign`
+ * call for a given `message`, or fed to `ecrecover` along with a signature to recover the public key
+ * used to produce the signature.
+ */
+export const hashPersonalMessage = function(message: any): Buffer {
+  const prefix = toBuffer(`\u0019Ethereum Signed Message:\n${message.length.toString()}`)
+  return keccak(Buffer.concat([prefix, message]))
+}
```
holgerd77: All fits well except for the last function ``hashPersonalMessage`` I would have a tendency to put this to signature related functionality.

src/pubkey.ts
```diff
@@ -0,0 +1,209 @@
...
+  return true
+}
+
+function calculateSigRecovery(v: number, chainId?: number): number {
+  return chainId ? v - (2 * chainId + 35) : v - 27
+}
+
+function isValidSigRecovery(recovery: number): boolean {
+  return recovery === 0 || recovery === 1
+}
```
holgerd77: I would have a tendency to move to top-most functions to ``address.js`` and make for the others some dedicated ``signature.js`` file.

src/pubkey.ts
```diff
@@ -0,0 +1,209 @@
...
+  r: Buffer
+  s: Buffer
+}
+
+/**
+ * Checks if the private key satisfies the rules of the curve secp256k1.
+ */
+export const isValidPrivate = function(privateKey: Buffer): boolean {
+  return secp256k1.privateKeyVerify(privateKey)
+}
```
holgerd77: One side note: on refactoring we should go through all the file names, I think there are better versions for many of them (like this one, but also ``fromSigned``, ``baToJSON``, ``zeros``) being more descriptive and generally better grasp what the function is doing. Think this would help a lot on code readability on the usage side where people (we 😄) are integrating this within their code.

s1na:
Applied your comments, the main differences are:

- Most of **common** renamed to **buffer**
- `defineProperties` moved to own file **object** (and marked as deprecated)
- Signature-related functions moved to **signature**
- Public key private key functions moved together with address related functions to a new file **account** (instead of **address**, since I think account is more general)

holgerd77:
Ok, will leave this open until tomorrow so that others have a chance to comment, but I think we've already got some good ground here.

src/account.ts
```diff
@@ -0,0 +1,180 @@
...
+    // Convert to SEC1 for secp256k1
+    return secp256k1.publicKeyVerify(Buffer.concat([Buffer.from([4]), publicKey]))
+  }
+
+  if (!sanitize) {
+    return false
+  }
+
+  return secp256k1.publicKeyVerify(publicKey)
+}
```
holgerd77: With these two I am unsure, if they rather belong in the `signature` area or here. Just did some context research, both are [used in wallet](https://github.com/ethereumjs/ethereumjs-wallet/blob/830307cb1b4d083a469e796e2c34064a6c948479/src/index.js#L24), I also had some short look at [this](https://hackernoon.com/a-closer-look-at-ethereum-signatures-5784c14abecc) HackerNoon article on Ethereum signature.

Think we can leave here for now, since we are not doing a release yet we can still move if found to be more appropriate in `signature`.

src/buffer.ts
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * Adds "0x" to a given `String` if it does not already start with "0x".
+ */
+export const addHexPrefix = function(str: string): string {
+  if (typeof str !== 'string') {
+    return str
+  }
+
+  return ethjsUtil.isHexPrefixed(str) ? str : '0x' + str
+}
```
holgerd77: Ok, this is not fitting at all. Should we call the whole file ``byte.js``? Then this is generally more generic and would also fit other functionality in the future eventually.

s1na:
Renamed buffer to bytes. I was thinking for future maybe it'd make sense to define custom types, and organize part of the code around those types. E.g. a type `Address` with various constructors (static methods) such as `zero()`, `create()`, `create2()`, `fromPublicKey()`, and instance methods like `toString()`, `toBuffer()`, `isValid()`.

holgerd77:
Hi Sina, I was out for a longer (baby-related 🍼 😄 ) walk and actually also thinking about how to better and more coherently impose hex value validation, your suggestion with custom types sounds great! I think I'll also leave function naming "as is" for this round if we might restructure in a bigger way mid-term, we should nevertheless do the API changes on this round.

Not sure how to organize some transition process, that's probably something for a second round and mid-term? Or would it already make sense to introduce parts of the functionality already here in parallel or something?

Anyhow. Will merge this PR for now. Can further brainstorm. 😄

holgerd77:
(or if you've got a strong opinion on this please let me know)

s1na:
> Or would it already make sense to introduce parts of the functionality already here in parallel or something?

Maybe we can introduce the types *in addition* to the existing API, test it in smaller libraries and only if it worked well do a full transition?

> can we do "byte.ts" instead of "bytes.ts"?

I don't have a strong opinion, and would happily change, but it does seem that in this case since we're always working with a collection of bytes and almost never with a single byte, `bytes` might convey the meaning better (similar packages I've seen have also been called bytes, e.g. go's [bytes](https://golang.org/pkg/bytes/)).

holgerd77:
Ok, let's keep, have merged.

holgerd77:
Makes sense (transition idea/suggestion).

PR #189, opened 2019-04-18
Remove the Standard banner

alcuadrado:
This project now uses prettier. The banner was from a previous setup.


holgerd77:
Do we want to replace this with a badge for prettier: https://github.com/prettier/prettier#badge

Side note: will soon do apply-on-all-libraries PR round, will include this change here with this PR as a reference change.

Issue #172, opened 2019-01-14
Breaking API changes (stricter type requirements)


holgerd77:
Currently some overhaul of the API with breaking API changes with stricter type requirements is [discussed](https://github.com/ethereumjs/ethereumjs-util/pull/170#issuecomment-452241863) within the `TypeScript` transition PR, will open this issue here for transparency, everyone interested feel free to join the discussion!

See API guidelines and change proposals in the thread.

holgerd77:
@alcuadrado @s1na I think it might be more targeted if we just come up with some comprise set of general API guidelines and rules (like "A string representing a hex value has always to be prefixed with ``0x``", similar rules, error case definitions) and we then go along the methods and modify so that these comply with the rules.

I will try to come up with a suggestion during the next 1-2 days, feel free to also already contribute, eventually we use a shared comment preceding this one and edit until we are satisfied.

We can eventually even add this to the organizational docs at the end and have some common guidelines for API creation, we could also point to on PRs and also use for self-guidance on refactoring and general library work.

holgerd77:
### API Guidelines and Rules DRAFT (Shared Post)

- A string representing a hex value has always to be prefixed with 0x

``TODO``

holgerd77:
@s1na @alcuadrado Ok, have done some first write-up of mainly hex and buffer-related API rules, which should be enough for a start and some discussion. Let me know what you think, is this too strict? You can also directly edit in the comment and change whatever you like.

Below I will paste the API change proposals from the ``TypeScript`` PR (linked above). Will try to adopt a bit to the latest state of discussion. Same here, feel free to edit directly.

holgerd77:
[WORK-IN-PROGRESS, READ IN DOCUMENT]

## Breaking API Changes

List with suggestions for breaking API changes

---

### Type: Bufferable
Should accept `HexString` type instead of `string`  if we agree on this.

### zeroAddress
Should also return `HexString` type if we agree on this.

### setLengthLeft / setLengthLeft
Also just did some organization-wide [search](https://github.com/search?q=org%3Aethereumjs+setLengthLeft&type=Code), this is in (EthereumJS) practice also mostly used on `Buffer` anyhow (doesn't mean there are other usages out there).

Would suggest that we drop the `Bufferable` (love the naming 😄) support here and just return `Buffer` here (respectively change the docs / return type to reality).

[...]

PR #197, opened 2019-05-09
toBuffer now throws if strings aren't 0x-prefixed hex values

alcuadrado:
I made this in response to the conversation we had in [this other PR](https://github.com/ethereumjs/ethereumjs-tx/pull/144).

As we agreed three, `toBuffer` had a surprising and error-prone behavior when a string other than 0x-prefixed hex one was given. This PR removes that behavior, throwing a clear error instead.

There were a few tests using this older behavior (i.e. passing utf-8 string), but that was not necessary and they were harder to read, so I replaced with equivalent 0x-prefixed hex strings.

`keccak` was also using and exposing this behavior, which makes sense as people sometimes hash strings. I implemented it inside `keccack` and documented it.

Finally, `hashPersonalMessage` was using it internally, but not exposing it, so I just updated it to an equivalent lower-level implementation.

Technically, this PR introduces a breaking change in `toBuffer`, but I think virtually all of the uses of `toBuffer` with non-hex strings were errors, as its behavior was unexpected and undocumented.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.idea/
+
 ### App ###
 
 .cachedb
--- a/docs/README.md
+++ b/docs/README.md
@@ -165,7 +165,7 @@ ___
 
 **● publicToAddress**: *[pubToAddress]()* =  pubToAddress
 
-*Defined in [index.ts:271](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/index.ts#L271)*
+*Defined in [account.ts:151](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/account.ts#L151)*
 
 ___
 <a id="setlength"></a>
@@ -174,7 +174,7 @@ ___
 
 **● setLength**: *[setLengthLeft]()* =  setLengthLeft
 
-*Defined in [index.ts:79](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/index.ts#L79)*
+*Defined in [bytes.ts:37](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/bytes.ts#L37)*
 
 ___
 <a id="stripzeros"></a>
@@ -183,7 +183,7 @@ ___
 
 **● stripZeros**: *[unpad]()* =  unpad
 
-*Defined in [index.ts:106](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/index.ts#L106)*
+*Defined in [bytes.ts:64](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/bytes.ts#L64)*
 
 ___
 
@@ -195,7 +195,7 @@ ___
 
 ▸ **addHexPrefix**(str: *`string`*): `string`
 
-*Defined in [index.ts:488](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/index.ts#L488)*
+*Defined in [bytes.ts:135](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/bytes.ts#L135)*
 
 Adds "0x" to a given `String` if it does not already start with "0x".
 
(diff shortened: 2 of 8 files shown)


alcuadrado:
Travis is failing with node 6 because the autoformat sometimes adds commas after the last argument in a function call. This is not supported in node 6.

This is an example of that pattern:
```js
assert.equal(
      ethUtils.MAX_INTEGER.toString('hex'),
      'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff',
    )
```

Are we still supporting node 6? It's not maintained anymore.

holgerd77:
Thanks for the PR, had a first look but need some more time to answer.

One first thing: can you please generally don't reformat code and remove the commit with the format changes in the tests file?

Some broader first note: maybe we want to take this occasion and generally do the API overhaul discussed here https://github.com/ethereumjs/ethereumjs-util/issues/172, eventually you already want to have a look and give this your own thoughts. This goes a bit in the direction of this PR, things have to be re-read a bit since we won't use a special HexString class for now which I proposed there.

@s1na do we eventually do this API overhaul a two-step process and first start with updates similar to the one being done in this PR and do integrations of your new types like the Address type https://github.com/ethereumjs/ethereumjs-util/pull/186 as a second step on top?

s1na:
I personally agree with this approach (and probably would go further and limit the APIs to only `Buffer` and throw an error otherwise :smile: ).

> @s1na do we eventually do this API overhaul a two-step process and first start with updates similar to the one being done in this PR and do integrations of your new types like the Address type #186 as a second step on top?

Yeah, I think that makes sense. They can even be possibly worked on in parallel as there's no conflict in between.

alcuadrado:
> One first thing: can you please generally don't reformat code and remove the commit with the format changes in the tests file?

Sure! I thought the idea was to always reformat everything with `ethereumjs-config-format-fix` and `ethereumjs-config-lint-fix`. Should tests be excluded from this?  Or how is this handled?

I updated it.

> Some broader first note: maybe we want to take this occasion and generally do the API overhaul discussed here #172, eventually you already want to have a look and give this your own thoughts. This goes a bit in the direction of this PR, things have to be re-read a bit since we won't use a special HexString class for now which I proposed there.

I'll go through that issue. Thanks for the link.


> I personally agree with this approach (and probably would go further and limit the APIs to only `Buffer` and throw an error otherwise 😄 ).

I'd love this, but wouldn't that mean breaking changes in all of ethereumjs?

holgerd77:
@alcuadrado With ``EthereumJS`` we are generally at the point where we can't have substantial progress to a greater extend any more if we don't do bigger steps and more significant changes on various libraries. We shouldn't do breaking stuff just for the change/breaking 😋, but if there is significant reasoning to do so we should absolutely do so to prepare some more solid ground for the future.

The ``ethereumjs-util`` library e.g. is definitely a candidate for some bigger changes especially on the API front. With dependency updates this is not too much of a problem. There is no need to do everything at once and we can just update "on occasion", we are already doing this regularly with things like that and that is the normal work progress work-flow to slowly but steadily evolve on all fronts.

holgerd77:
@alcuadrado can you please rebase your branch for a review?

alcuadrado:
> @alcuadrado With `EthereumJS` we are generally at the point where we can't have substantial progress to a greater extend any more if we don't do bigger steps and more significant changes on various libraries. We shouldn't do breaking stuff just for the change/breaking 😋, but if there is significant reasoning to do so we should absolutely do so to prepare some more solid ground for the future.
> 
> The `ethereumjs-util` library e.g. is definitely a candidate for some bigger changes especially on the API front. With dependency updates this is not too much of a problem. There is no need to do everything at once and we can just update "on occasion", we are already doing this regularly with things like that and that is the normal work progress work-flow to slowly but steadily evolve on all fronts.

Then, I agree with using `Buffer` everywhere. That'd also make much easier to type most of the things that now have `any`. Let's continue this discussion in #172 :)

> @alcuadrado can you please rebase your branch for a review?

Done

PR #204, opened 2019-06-24
Implement EIP1191

### LINKED ISSUE #201: EIP-1191: new address checksum algorithm
[This EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1191.md) introduces a new checksum algorithm, which adds the `chainId` to the checksum calculation.

This new checksum has already been implemented in web3.js, trezor, ledger, mycrypto, and myetherwallet. Some of them are already using it for RSK, as it was originally proposed as [RSKIP-60](https://github.com/rsksmart/RSKIPs/blob/master/IPs/RSKIP60.md).

We should probably implement it too. What do you think? @s1na @holgerd77 

This can be done in a backward compatible way as if no `chainId` is used, it defaults to EIP-55 checksums.

alcuadrado:
This PR implements [EIP-1191 address checksum algorithm](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1191.md).

Fixes #201

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -32,13 +32,17 @@ export const isZeroAddress = function(address: string): boolean {
 /**
  * Returns a checksummed address.
  *
- * If a chainId is provided, the result will be an EIP-1191 checksum. Otherwise, it will be an
- * EIP-55 checksum.
+ * If a eip1191ChainId is provided, the chainId will be included in the checksum calculation. This
+ * has the effect of checksummed addresses for one chain having invalid checksums for others.
+ * For more details, consult EIP-1191.
+ *
+ * WARNING: Checksums with and without the chainId will differ. As of 2019-06-26, the most commonly
+ * used variation in Ethereum was without the chainId. This may change in the future.
  */
-export const toChecksumAddress = function(address: string, chainId?: number): string {
+export const toChecksumAddress = function(address: string, eip1191ChainId?: number): string {
   address = ethjsUtil.stripHexPrefix(address).toLowerCase()
 
-  const prefix = chainId !== undefined ? chainId.toString() + '0x' : ''
+  const prefix = eip1191ChainId !== undefined ? eip1191ChainId.toString() + '0x' : ''
 
   const hash = keccak(prefix + address).toString('hex')
   let ret = '0x'
@@ -57,11 +61,10 @@ export const toChecksumAddress = function(address: string, chainId?: number): st
 /**
  * Checks if the address is a valid checksummed address.
  *
- * If a chainId is provided, the address is compared to an EIP-1191 checksum. Otherwise, it will be
- * compared to an EIP-55 checksum.
+ * See toChecksumAddress' documentation for details about the eip1191ChainId parameter.
  */
-export const isValidChecksumAddress = function(address: string, chainId?: number): boolean {
-  return isValidAddress(address) && toChecksumAddress(address, chainId) === address
+export const isValidChecksumAddress = function(address: string, eip1191ChainId?: number): boolean {
+  return isValidAddress(address) && toChecksumAddress(address, eip1191ChainId) === address
 }
 
 /**
--- a/src/account.ts
+++ b/src/account.ts
@@ -31,10 +31,16 @@ export const isZeroAddress = function(address: string): boolean {
 
 /**
  * Returns a checksummed address.
+ *
+ * If a chainId is provided, the result will be an EIP-1191 checksum. Otherwise, it will be an
+ * EIP-55 checksum.
  */
-export const toChecksumAddress = function(address: string): string {
+export const toChecksumAddress = function(address: string, chainId?: number): string {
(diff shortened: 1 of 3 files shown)


alcuadrado:
I removed node 6 from the travis config, as a test was failing because of that. 

I'm pretty sure we don't support node 6 anymore as its EOL has already happened. Let me know if i'm wrong and I'll revert that change.

src/account.ts
```diff
@@ -31,10 +31,16 @@ export const isZeroAddress = function(address: string): boolean {
...
+ *
+ * If a chainId is provided, the result will be an EIP-1191 checksum. Otherwise, it will be an
+ * EIP-55 checksum.
  */
-export const toChecksumAddress = function(address: string): string {
+export const toChecksumAddress = function(address: string, chainId?: number): string {
   address = ethjsUtil.stripHexPrefix(address).toLowerCase()
-  const hash = keccak(address).toString('hex')
+
+  const prefix = chainId !== undefined ? chainId.toString() + '0x' : ''
```
s1na: The documentation makes it clear, but if someone seeing `chainId` as the second parameter provides an Ethereum chainId (e.g. `1`) then the checksum will fail.

One alternative is to provide a second function `toEIP1991ChecksumAddress` (maybe with a better name :smile:).

I don't have a strong opinon however. @holgerd77 what do you think?

src/account.ts
```diff
@@ -50,9 +56,12 @@ export const toChecksumAddress = function(address: string): string {
 
 /**
  * Checks if the address is a valid checksummed address.
+ *
+ * If a chainId is provided, the address is compared to an EIP-1191 checksum. Otherwise, it will be
```
s1na: Can you also re-generate the docs?

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: node_js
 node_js:
-  - "6"
```
s1na: I merged #207, this is not necessary anymore

test/index.js
```diff
@@ -656,23 +656,103 @@ const checksumAddresses = [
...
+        for (const addr of addresses) {
+          assert.equal(ethUtils.isValidChecksumAddress(addr, chainId), true)
+        }
+      }
+    })
+
+    it("Should return false for invalid cases", function () {
+      // If we set the chain id, an EIP55 encoded address should be invalid
+      for (let i = 0; i < eip55ChecksumAddresses.length; i++) {
+        assert.equal(ethUtils.isValidChecksumAddress(eip55ChecksumAddresses[i], 1), false)
```
s1na: Ah yeah this is the test case I was thinking about :)

src/account.ts
alcuadrado: That's a good point.

I made it this way to keep it similar to other implementations, like [web3.js' one](https://github.com/ethereum/web3.js/blob/aaf26c8806bc9fb60cf6dcb6658104963c6c7fc7/packages/web3-utils/src/index.js#L261)

Another alternative is to rename the second param to something like `eip1191ChainId`

src/account.ts
alcuadrado: I thought they were regenerated when a new release is published

holgerd77: Hmm, we still don't have a clear policy here. Until now I had also some tendency to the "regenerate the docs once the code changes" variant. On a second thought the docs are primarily for the users using the releases, so probably the suggestion from @alcuadrado to generate on release makes more sense. @s1na what do you think?

src/account.ts
holgerd77: I would have a tendency to the renamed parameter. @alcuadrado can you also add "(prefixed with chainId)" and "(no prefix)" or some similar text after the respective EIP mentions in the JSDocs? Think this will also make things more clear, I often find these plain EIP mentions in our docs not very helpful if one is not totally in the EIP specs.

holgerd77: One can also add some "Be are that checksums will be different" or so sentence.

holgerd77: Are -> aware

src/account.ts
s1na: Just like you I had no clear approach in mind, but I think regenerating them on release does make sense. Let's go with that.

alcuadrado: TBH I'm not sure why I had that in mind, but I think one of its advantages is that PRs are smaller.

src/account.ts
alcuadrado: I changed the param name and updated the doc.

alcuadrado:
Travis got stuck, so I pushed the same code again. Can you reapprove @s1na? Thanks!

kumavis:
@holgerd77 @alcuadrado @s1na can we get a new version published? thanks/

holgerd77:
@kumavis Yes - sorry, wasn't aware that actually didn't yet release all the latest changes for such a long time. Will prepare something within the next days, hopefully this, eventually next week.

Issue #198, opened 2019-05-13
Wasm modules for crypto


s1na:
This is a spin-off discussion from #195 which aims to assess whether it makes sense to use Wasm modules for crypto.

The arguments for this approach being that WebAssembly is now available everywhere (node & browsers) in comparison to `node-gyp` bindings (used by the `keccak` and `secp256k1`) which is (as the name suggests) only available in `node` (in browser a native JS code is executed).

What remains to be seen is how the performance of an efficient Wasm module compares against the `node-gyp` bindings and native JS code.

s1na:
I did a simple benchmark (not sure about accuracy) for `keccak`:

- `keccak` lib (with `node-gyp` binding) hovers around 20-30 us per hash computation
- `keccak/js` is around 150-160 us
- [keccak256_rhash.wasm](https://github.com/poemm/C_ewasm_contracts/blob/master/wasm/keccak256_rhash.wasm) ewasm precompile is around 260-280k us

The precompile result also includes instantiation of the Wasm module and could potentially have overhead due to calling host functions for fetching input data, as seen in the source:

https://github.com/poemm/C_ewasm_contracts/blob/master/src/keccak256_rhash.c#L509-L521

I'll next try a simpler Wasm module (which doesn't call host functions).

holgerd77:
👍

s1na:
Slightly modified [keccak256_rhash.wasm](https://github.com/s1na/C_ewasm_contracts/commit/450fc5d1aef4c3444fbb9aa4a096e6fbce26cca4#diff-29b402012a429a1240b74ed6cfd0186e) computes hashes in 10-20 us, which is slightly faster even than the native `node-gyp` binding. It's currently limited to inputs less than 128kb (which should be fine for most use-cases), but this can be fixed. So far I've only tested really simple inputs, I'll have to test more.

The gain comes from two sides:

- The wasm module now takes its input and a pointer to output directly as arguments of the `_main` function which avoids multiple EEI calls (downside being it's not ewasm-compatible anymore)
- Instantiation of the module constituted most of the time taken. Thanks to immense help from @poemm, we managed to avoid re-initializing the module every time (by zeroing out relevant parts of memory and avoiding a mutable global variable). This is the js [code](https://gist.github.com/s1na/92eae826c696de5e4d60708f8743e59f) for calling the wasm module.

holgerd77:
Oh, that sounds very promising, great!

holgerd77:
[...]

PR #228, opened 2020-01-29
Update secp2561 ECDSA dependency to v4.0.0

holgerd77:
Did some first tests here, as @alcuadrado stated the API changed significantly and there are currently 11 tests failing for functionality in `account.ts` and `signature.ts`.

Main thing here is that the API changed from using `Buffer` values to `Uint8Array` values, see [v4 API docs](https://github.com/cryptocoinjs/secp256k1-node/blob/master/API.md) and [v3 API docs](https://github.com/cryptocoinjs/secp256k1-node/blob/v3.x/API.md) for comparison.

Two things to address here:

### 1. Analyze API Change Implications

My knowledge of the implications of this switch is actually limited. I would need some additional voices here from people knowing better what this comes along with on the performance and compatibility side.

### 2. Library Fixes

I would assume that (at least on a first round) it would make most sense to wrap the `Uint8Array` usage and hide this from users of this library, since the whole EthereumJS ecosystem is based on `Buffer` usage anyhow?

Did a first test wrapping the `isValidPrivate` function, this didn't improve the test success numbers, not sure why, this would need some further analysis.

```typescript
/**
 * Checks if the private key satisfies the rules of the curve secp256k1.
 */
export const isValidPrivate = function(privateKey: Buffer): boolean {
  const privateKeyU = new Uint8Array(privateKey)
  console.log(privateKey)
  console.log(privateKeyU)
  return secp256k1.privateKeyVerify(privateKeyU)
}
```

Generally there would minimally be the need to get all test passing again.

// @michaelsbradleyjr @alcuadrado @iurimatias @evertonfraga @ryanio 

Follow-up on #225

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "ethjs-util": "0.1.6",
     "keccak": "^3.0.0",
     "rlp": "^2.2.4",
-    "secp256k1": "^4.0.0"
+    "secp256k1": "^4.0.1"
   },
   "devDependencies": {
     "@ethereumjs/config-prettier": "^1.1.0",
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "ethjs-util": "0.1.6",
     "keccak": "^3.0.0",
     "rlp": "^2.2.4",
-    "secp256k1": "^3.0.1"
+    "secp256k1": "^4.0.0"
   },
   "devDependencies": {
     "@ethereumjs/config-prettier": "^1.1.0",
--- a/package.json
+++ b/package.json
@@ -100,7 +100,7 @@
     "@ethereumjs/config-tslint": "^1.1.0",
     "@types/mocha": "^5.2.7",
     "@types/node": "^11.9.0",
-    "@types/secp256k1": "3.5.0",
+    "@types/secp256k1": "^4.0.1",
     "contributor": "^0.1.25",
     "husky": "^2.1.0",
     "karma": "^4.0.0",
--- a/src/account.ts
+++ b/src/account.ts
@@ -2,7 +2,7 @@ const ethjsUtil = require('ethjs-util')
 import * as assert from 'assert'
 import * as secp256k1 from 'secp256k1'
 import * as BN from 'bn.js'
-import { zeros, bufferToHex } from './bytes'
+import { zeros, bufferToHex, toBuffer } from './bytes'
 import { keccak, keccak256, rlphash } from './hash'
 import { assertIsHexString, assertIsBuffer } from './helpers'
 
@@ -129,6 +129,7 @@ export const isValidPrivate = function(privateKey: Buffer): boolean {
  * @param sanitize Accept public keys in other formats
  */
 export const isValidPublic = function(publicKey: Buffer, sanitize: boolean = false): boolean {
+  assertIsBuffer(publicKey)
   if (publicKey.length === 64) {
     // Convert to SEC1 for secp256k1
     return secp256k1.publicKeyVerify(Buffer.concat([Buffer.from([4]), publicKey]))
--- a/src/account.ts
+++ b/src/account.ts
@@ -122,7 +122,18 @@ export const generateAddress2 = function(
  * Checks if the private key satisfies the rules of the curve secp256k1.
  */
 export const isValidPrivate = function(privateKey: Buffer): boolean {
-  return secp256k1.privateKeyVerify(privateKey)
+  let valid
+  try {
+    valid = secp256k1.privateKeyVerify(privateKey)
+  } catch (e) {
+    // TODO: Change API to also throw on wrong input instead of returning false
+    if (e.message === 'Expected private key to be an Uint8Array with length 32') {
+      valid = false
+    } else {
+      throw e
+    }
+  }
+  return valid
 }
 
 /**
--- a/src/account.ts
+++ b/src/account.ts
@@ -122,18 +122,7 @@ export const generateAddress2 = function(
  * Checks if the private key satisfies the rules of the curve secp256k1.
  */
 export const isValidPrivate = function(privateKey: Buffer): boolean {
-  let valid
-  try {
-    valid = secp256k1.privateKeyVerify(privateKey)
-  } catch (e) {
-    // TODO: Change API to also throw on wrong input instead of returning false
-    if (e.message === 'Expected private key to be an Uint8Array with length 32') {
-      valid = false
-    } else {
-      throw e
-    }
-  }
-  return valid
+  return secp256k1.privateKeyVerify(privateKey)
 }
 
 /**
--- a/src/account.ts
+++ b/src/account.ts
@@ -164,7 +164,7 @@ export const isValidPublic = function(publicKey: Buffer, sanitize: boolean = fal
 export const pubToAddress = function(pubKey: Buffer, sanitize: boolean = false): Buffer {
   pubKey = toBuffer(pubKey)
   if (sanitize && pubKey.length !== 64) {
-    pubKey = secp256k1.publicKeyConvert(pubKey, false).slice(1)
+    pubKey = toBuffer(secp256k1.publicKeyConvert(pubKey, false).slice(1))
   }
   assert(pubKey.length === 64)
   // Only take the lower 160bits of the hash
@@ -187,7 +187,7 @@ export const privateToAddress = function(privateKey: Buffer): Buffer {
 export const privateToPublic = function(privateKey: Buffer): Buffer {
   privateKey = toBuffer(privateKey)
   // skip the type flag and use the X, Y points
-  return secp256k1.publicKeyCreate(privateKey, false).slice(1)
+  return toBuffer(secp256k1.publicKeyCreate(privateKey, false).slice(1))
 }
 
 /**
@@ -196,7 +196,7 @@ export const privateToPublic = function(privateKey: Buffer): Buffer {
 export const importPublic = function(publicKey: Buffer): Buffer {
   publicKey = toBuffer(publicKey)
   if (publicKey.length !== 64) {
-    publicKey = secp256k1.publicKeyConvert(publicKey, false).slice(1)
+    publicKey = toBuffer(secp256k1.publicKeyConvert(publicKey, false).slice(1))
   }
   return publicKey
 }
--- a/src/externals.ts
+++ b/src/externals.ts
@@ -6,7 +6,6 @@
  */
 
 const ethjsUtil = require('ethjs-util')
-import * as secp256k1 from 'secp256k1'
 import * as BN from 'bn.js'
 import * as rlp from 'rlp'
 
@@ -24,8 +23,3 @@ export { BN }
  * [`rlp`](https://github.com/ethereumjs/rlp)
  */
 export { rlp }
-
-/**
- * [`secp256k1`](https://github.com/cryptocoinjs/secp256k1-node/)
- */
-export { secp256k1 }
--- a/src/externals.ts
+++ b/src/externals.ts
@@ -1,7 +1,7 @@
 /**
  * Re-exports commonly used modules:
  * * Adds [`ethjs-util`](https://github.com/ethjs/ethjs-util) methods.
- * * Exports [`BN`](https://github.com/indutny/bn.js), [`rlp`](https://github.com/ethereumjs/rlp), [`secp256k1`](https://github.com/cryptocoinjs/secp256k1-node/).
+ * * Exports [`BN`](https://github.com/indutny/bn.js), [`rlp`](https://github.com/ethereumjs/rlp).
  * @packageDocumentation
  */
 
--- a/src/signature.ts
+++ b/src/signature.ts
@@ -17,12 +17,12 @@ export const ecsign = function(
   privateKey: Buffer,
   chainId?: number,
 ): ECDSASignature {
-  const sig = secp256k1.sign(msgHash, privateKey)
-  const recovery: number = sig.recovery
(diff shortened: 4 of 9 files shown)


holgerd77:
Have also submitted an issue #229 to split the test cases into separate files, this would actually ease the work here.

holgerd77:
Rebased this on top of #231

package.json
```diff
@@ -92,7 +92,7 @@
     "ethjs-util": "0.1.6",
     "keccak": "^3.0.0",
     "rlp": "^2.2.3",
-    "secp256k1": "^3.0.1"
+    "secp256k1": "^4.0.0"
```
michaelsbradleyjr: With this bump and keccak having been bumped to `^3.0.0` should an `"engines"` field be added to package.json?

```
"engines": { "node": ">=10.0.0" }
```

Likewise, should Node 8.x be dropped from the testing matrix?

holgerd77: I have no experience with the `engines` field but I would assume it would make sense to use this throughout the EthereumJS libraries in general. Does this prevent an installation from happening when defined too strict?

Yes, Node 8 should then also be dropped from the test matrix. We nevertheless need to get these basic decisions outlined above done first.

michaelsbradleyjr: `"engines"` behavior varies depending on whether `npm` is used or `yarn` is used.

In the case of `npm`, the restriction is only enforced with respect to *your project's* `package.json`. That is, if dev/Deps listed in your project's `package.json` have an `"engines"` setting that isn't satisfied, it doesn't matter, they'll get installed anyway. But if your project's `package.json` has an `"engines"` setting that isn't satisfied, then running `npm install` in your project will fail.

In the case of `yarn` both your project's and its dependencies' `"engines"` settings are checked and `yarn install` will fail if there is any mismatch.

Either way, the `"engines"` setting is helpful metadata for humans, signaling the package manager's intention to support some Node versions and not others.

holgerd77: Ok, thanks, but then I would assume at least to define a max version is somewhat dangerous and should eventually be skipped.

michaelsbradleyjr: Typically you only want a max version if you know that, for some reason, your package isn't compatible with a newer version of node.

For awhile the `embark` package had:

```
"engines": { "node": ">=10.17.0 <12.0.0" }
```

10.17.0 was the first Node version that had a change to Node's built-in scrypt that we wanted to be sure was in-place in a user's setup; at the same time, we had some transitive dependencies that weren't yet compatible with Node `>=12.0.0`. We eventually resolved the latter and now we have:

```
"engines": { "node": ">=10.17.0" }
```

holgerd77: Yeah, right, likely a case-by-case decision. If there is a dependency clearly not compatible on the upper bound it makes sense to narrow. If one is is potentially - and unnecessarily - preventing users to upgrade to a future Node version, this should be left open.

holgerd77:
Update: had another short look, actually it's not the success cases which fail (e.g. `isValidPrivate(Buffer.from('fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364140', 'hex'))` ("should work otherwise (< N)" test case in `test/account.spec.js`) still evaluates to `true`), it's the error cases.

These now throw with the new library (behavior before: evaluate to `false`), e.g. `isValidPrivate(Buffer.from('0011223344', 'hex'))` ("should fail on short input" test case):

```shell
PATH/ethereumjs-util/node_modules/secp256k1/lib/index.js:18
  if (!cond) throw new Error(msg)
             ^

Error: Expected private key to be an Uint8Array with length 32
```

So seems we have to decide here if we want to adopt here along the new API behavior (throw as well) or wrap this up in a `try/catch` and continue to just return `false`.

Likely also a cc for @alcuadrado 😄 

Maybe a two-step approach makes sense: on the first round wrap this up to be able to do a minor release here but keep this in mind to change the behavior on the next major release?

holgerd77:
Ok, I have now added an exemplary backwards-compatible fix for the `isValidPrivate()` functionality, preserving the old "wrong length input" -> return `false` behavior and re-throwing on other cases.

If no objection I would continue to do fixes like this towards a backwards-compatible release with an updated `secp2561` library. Further down the line I think the API should be changed here.

alcuadrado:
> My knowledge of the implications of this switch is actually limited. I would need some additional voices here from people knowing better what this comes along with on the performance and compatibility side.

I think there are no performance implications, `Uint8Array` is also just a byte array. I wouldn't be surprised if v8 better optimizes those.

Regarding compatibility, `Uint8Array` has been standard since 2015, so it's supported everywhere now. In an ideal world, everything would use `Uint8Array` instead of `Buffer`, as `Buffer` has to be polyfill in the browser.



> I would assume that (at least on a first round) it would make most sense to wrap the `Uint8Array` usage and hide this from users of this library, since the whole EthereumJS ecosystem is based on `Buffer` usage anyhow?

I agree. I'd do this to get a minor release out. I'd also do things like the `try`/`catch` you introduced to keep compatibility.

What worries me about this change, is that this library re-exports `secp256k1` here: https://github.com/ethereumjs/ethereumjs-util/blob/master/src/index.ts#L20 -- This makes upgrading a breaking change.

holgerd77:
> What worries me about this change, is that this library re-exports `secp256k1` here: https://github.com/ethereumjs/ethereumjs-util/blob/master/src/index.ts#L20 -- This makes upgrading a breaking change.

Thanks a lot for catching this, wouldn't have thought about this along the way! Thought about this a bit longer: I think we then just can't do the release in the intended way - so as a minor release - no way to get around that here without risking to break stuff for people.

Since we had this `minor` vs `major` release discussion already I would suggest that we will then finally target a `major` release. From my perspective it would then also make a lot of sense to just give this another - let's say four weeks - and use the occasion and to...

1. **Improve error handling** on the library (e.g. in the cases already touched here)
2. Do the **breaking API changes** discussed [here](https://github.com/ethereumjs/ethereumjs-util/issues/172#issuecomment-491756771) (this list is likely not up-to-date and needs some modifications) and take the chance and make the API more strict and well-defined regarding the input values.

@alcuadrado Does this make sense? I know people are eager to get this released but I think this would be very much worth it the short additional waiting period.

Will also include @s1na here to take notice (and eventually also leave a comment or suggestion).

alcuadrado:
Makes sense @holgerd77.

I think I'd also remove the re-export of `secp256k1` if we are introducing a breaking change around it. This way we'll avoid repeating this situation (i.e. being forced to a breaking change by it) in the future.

TBH, I'm not sure what's the advantage of that export anyway.

cgewecke:
There's more to do here...just leaving a couple intermediate notes so things don't get forgotten.

+ the Node 8 job passes but it's technically no longer supported per [secp256k1 160][2]. 
  + Could be removed?   

+ **There is a behavioral change** with validation of invalid SEC1 public keys **in the browser (only)**.. This is shown by:
    + [Failing karma test][1] in cc300dd
    + Tests passing with 4dd6580

   It looks like invalid SEC1 keys will validate in the browser if sanitization is turned on. There 
   is a recent change in secp's browser fallback [elliptic library][3]...not certain what to make of this 
   tbh...
 

[1]: https://github.com/ethereumjs/ethereumjs-util/pull/228/checks?check_run_id=610836864#step:5:277
[2]: https://github.com/cryptocoinjs/secp256k1-node/pull/160#issue-358514532
[3]: https://github.com/cryptocoinjs/secp256k1-node/blob/master/lib/elliptic.js

cgewecke:
RE: Invalid SEC1 tests. 

Secpk256k1's tests for publicKeyVerify [are here][1] and look like this:
```javascript
    t.test('validate invalid public keys', (t) => {
      const privateKey = util.getPrivateKey()
      const publicKey = util.getPublicKey(privateKey)

      const invalidVersion = Buffer.from(publicKey.compressed)
      invalidVersion[0] = 0x00
      t.false(secp256k1.publicKeyVerify(invalidVersion), 'invalid version byte')

      const invalidY = Buffer.from(publicKey.uncompressed)
      invalidY[64] ^= 0x01
      t.false(secp256k1.publicKeyVerify(invalidY), 'invalid Y')

      t.end()
    })
```

They prefix the array with `0x00`. The invalid SEC1 key in the ejs-utils suite looks like this:

https://github.com/ethereumjs/ethereumjs-util/blob/3bbc948a1887df5f10fb0ebc3f07b7b9f774302c/test/account.spec.ts#L148 

Changing our test case to so that it resembles secp256k1's causes the tests to pass as expected (e.g without making the options change).

But this doesn't explain why the result for the existing case is different in the browser vs. node...think we may need to open an issue asking for advice at secp256k1 unless someone already knows what's going on here. 

[1]: https://github.com/cryptocoinjs/secp256k1-node/blob/d9f7bf09a4c6f9ffd91615adf9c6d2b301107fa4/test/publickey.js#L19-L32

holgerd77:
@cgewecke Can you do a branch update here?

holgerd77:
> * the Node 8 job passes but it's no longer supported per [secp256k1 160](https://github.com/cryptocoinjs/secp256k1-node/pull/160#issue-358514532). Could be removed?

Yeah, Node 8 can be dropped from the test matrix

> But this doesn't explain why the result for the existing case is different in the browser vs. node...think we may need to open an issue asking for advice at secp256k1 unless someone already knows what's going on here.

Yes, might be really worth to open an issue over there, can you do that? Maybe we are lucky and get a short-term answer/investigation.

cgewecke:
@holgerd77 
> Yeah, Node 8 can be dropped from the test matrix
👍 

> Yes, might be really worth to open an issue over there, can you do that? Maybe we are lucky and get a short-term answer/investigation.

Yes I will make a minimal reproduction case for that today and report at secp256k1. Have looked through the issues at elliptic as well and haven't seen anything that looks exactly like this problem...

holgerd77:
Just reread some issues and stumbled over the "WASM modules for Crypto" issue https://github.com/ethereumjs/ethereumjs-util/issues/198 opened by @s1na.

If this here causes some problems I wonder if it might be more promising (and - apart from that - generally more future-proof eventually) to play around with https://github.com/bitauth/bitcoin-ts mentioned in the thread. @cgewecke what do you think?

cgewecke:
@holgerd77 One thing about `bitauth/bitcoin-ts` is that it [uses BigInt][1]. @ricmoo has identified some drawbacks with this while reviewing a similar proposal in [ethers 727][2] - they're not supported by Safari and therefore can't be used on iOS. He has decided to defer BigInt support until Ethers 6. 

New things in this PR since last look...
+ rebased, removed Node 8, and did a little cleanup. 
+ opened [secp256k1 166][3] asking for advice about the invalid SEC1 tests.

[1]: https://github.com/bitauth/bitcoin-ts/blob/35aac14bba9b4fa5b4d01f10a2308a6836effd0e/src/lib/format/numbers.ts#L144
[2]: https://github.com/ethers-io/ethers.js/pull/727#issuecomment-583347250
[3]: https://github.com/cryptocoinjs/secp256k1-node/issues/166

holgerd77:
`bitauth/bitcoin-ts`: ah ok, wasn't aware this is using `BigInt`, than this is also likely something being 1-2 more major releases away on our side.

Thanks for opening the issue over on the `secp2561-node` library, @fanatid has already reacted and released [v4.0.1](https://github.com/cryptocoinjs/secp256k1-node/issues/166) (so great, thanks for that! 🎉 🙂), @cgewecke than you can directly update the dependency and we should be ready to go here?

Have removed the Node 8 tests from the checks being required, so this is now passing CI. Let me know once this is ready for review.

package.json
holgerd77: @cgewecke ah, and can you also add an `engines` field with `"engines": { "node": ">=10.0.0" }` as suggested in the first comment here?

cgewecke:
@holgerd77 040cb54 updates to 4.0.1 and reverts the SEC1 test case changes. Looks like everything's working ok now.

src/account.ts
```diff
@@ -182,7 +183,7 @@ export const privateToPublic = function(privateKey: Buffer): Buffer {
 export const importPublic = function(publicKey: Buffer): Buffer {
   assertIsBuffer(publicKey)
   if (publicKey.length !== 64) {
-    publicKey = secp256k1.publicKeyConvert(publicKey, false).slice(1)
+    publicKey = toBuffer(secp256k1.publicKeyConvert(publicKey, false).slice(1))
   }
   return publicKey
 }
```
holgerd77, line 189: Ok, have checked for all `return` value conversions, looks complete.

src/externals.ts
```diff
@@ -24,8 +23,3 @@ export { BN }
  * [`rlp`](https://github.com/ethereumjs/rlp)
  */
 export { rlp }
-
-/**
- * [`secp256k1`](https://github.com/cryptocoinjs/secp256k1-node/)
- */
-export { secp256k1 }
```
holgerd77, line 31: For reference: removal of the `secp2561` re-export

src/signature.ts
```diff
@@ -45,8 +45,8 @@ export const ecrecover = function(
   if (!isValidSigRecovery(recovery)) {
     throw new Error('Invalid signature v value')
   }
-  const senderPubKey = secp256k1.recover(msgHash, signature, recovery)
-  return secp256k1.publicKeyConvert(senderPubKey, false).slice(1)
+  const senderPubKey = secp256k1.ecdsaRecover(signature, recovery, msgHash)
+  return toBuffer(secp256k1.publicKeyConvert(senderPubKey, false).slice(1))
 }
 
 /**
```
holgerd77, line 52: Ok, also checked for completeness here.

holgerd77:
This has now been published in the [v7.0.0](https://github.com/ethereumjs/ethereumjs-util/releases/tag/v7.0.0) release on npm.

PR #235, opened 2020-02-04
Add basic API tests for re-exports

### LINKED ISSUE #234: Add basic API tests for re-exports
Along with an update of the version of the `secp2561` dependency https://github.com/ethereumjs/ethereumjs-util/pull/228 we were close to updating a re-export with a breaking version within a minor `ethereumjs-util` release, see the [comment](https://github.com/ethereumjs/ethereumjs-util/pull/228) from @alcuadrado for context.

To have this more structurally addressed and not only rely on reviewer awareness we should add some test cases running some API functionality of the re-exports (BN.js, RLP, secp2561). I would suggest to start slow here and add 3-5 test cases of central functionality of the libraries (we can't realistically test the whole API) for each re-export to the new `test/externals.spec.ts` test file.

This won't be the everything-covered solution but already significantly enlarge the chance to catch structural changes like a reworked error handling or a Node version dropped in the re-export library.

I would suggest (first thought, open for improvements along the way) to do at least the following tests for each library:

- Use a function as intended
- Trigger a well-defined error case on a function
- Eventually trigger a not-so-well defined error case (wrong input)

The error case differentiation is a bit blurry, might be best to decide on a case-by-case basis what is practical.

ryanio:
This PR adds some basic tests for the re-exports in the new test file `externals.spec.ts`:

```
  External BN export
    ✓ should export `BN`
    ✓ should use a BN function correctly
    ✓ should throw on exceptions

  External rlp export
    ✓ should export `rlp`
    ✓ should use a rlp function correctly
    ✓ should throw on exceptions

  External secp256k1 export
    ✓ should export `secp256k1`
    ✓ should use a secp256k1 function correctly
    ✓ should throw on exceptions

  External ethjsUtil export
    ✓ should have all ethjsUtil methods
    ✓ should use ethjsUtil functions correctly
    ✓ should handle exceptions and invalid inputs
```

Closes #234.

### DIFF
--- a/src/object.ts
+++ b/src/object.ts
@@ -14,7 +14,7 @@ import { toBuffer, baToJSON, stripZeros } from './bytes'
  * @param data data to be validated against the definitions
  * @deprecated
  */
-export const defineProperties = function(self: any, fields: any, data: any) {
+export const defineProperties = function(self: any, fields: any, data?: any) {
   self.raw = []
   self._fields = []
 
--- a/test/account.spec.ts
+++ b/test/account.spec.ts
@@ -1,5 +1,6 @@
-import assert = require('assert')
-const {
+import * as assert from 'assert'
+import * as BN from 'bn.js'
+import {
   isValidPrivate,
   isValidPublic,
   importPublic,
@@ -13,8 +14,7 @@ const {
   isValidChecksumAddress,
   isValidAddress,
   toChecksumAddress,
-} = require('../src')
-import BN = require('bn.js')
+} from '../src'
 const eip1014Testdata = require('./testdata/eip1014Examples.json')
 
 describe('isValidPrivate', function() {
--- a/test/bytes.spec.ts
+++ b/test/bytes.spec.ts
@@ -1,5 +1,6 @@
 import * as assert from 'assert'
-const {
+import * as BN from 'bn.js'
+import {
   zeros,
   zeroAddress,
   isZeroAddress,
@@ -8,16 +9,13 @@ const {
   setLengthLeft,
   setLengthRight,
   bufferToHex,
-  intToHex,
-  intToBuffer,
   bufferToInt,
   fromSigned,
   toUnsigned,
   addHexPrefix,
   toBuffer,
   baToJSON,
-} = require('../src')
-import BN = require('bn.js')
+} from '../src'
 
 describe('zeros function', function() {
   it('should produce lots of 0s', function() {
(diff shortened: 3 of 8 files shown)


src/object.ts
```diff
@@ -14,7 +14,7 @@ import { toBuffer, baToJSON, stripZeros } from './bytes'
  * @param data data to be validated against the definitions
  * @deprecated
  */
-export const defineProperties = function(self: any, fields: any, data: any) {
+export const defineProperties = function(self: any, fields: any, data?: any) {
```
holgerd77: For reference: if I understand correctly this is a correction of the API regarding the currently existing optionality of the `data` param within the function code. It can be regarded as a bug fix, but should be pointed out in the next release notes.

test/account.spec.ts
```diff
@@ -13,8 +14,7 @@ const {
   isValidChecksumAddress,
   isValidAddress,
   toChecksumAddress,
-} = require('../src')
-import BN = require('bn.js')
+} from '../src'
```
holgerd77: Some unrelated import improvements, nice, and justifiable since only touching the tests.

test/bytes.spec.ts
```diff
@@ -109,22 +107,6 @@ describe('bufferToHex', function() {
...
-})
-
-describe('intToBuffer', function() {
-  it('should convert a int to a buffer', function() {
-    const i = 6003400
-    const buf = intToBuffer(i)
-    assert.equal(buf.toString('hex'), '5b9ac8')
-  })
-})
-
```
holgerd77: These tests moved to the `externals.spec.ts` file, ok.

test/externals.spec.ts
```diff
@@ -1,29 +1,167 @@
...
+  const num = Math.pow(2, 53)
+  assert.throws(() => {
+    return new src.BN(num, 10)
+  }, /^Error: Assertion failed$/)
+
+  // should throw error with num eq 0x4000000
+  assert.throws(function() {
+    new src.BN(0).iaddn(0x4000000)
+  }, /^Error: Assertion failed$/)
+})
```
holgerd77: Nice selection, ok.

For my understanding: what is this 0x4000000 case?

test/externals.spec.ts
```diff
@@ -1,29 +1,167 @@
...
+      'hex',
+    )
+
+    let res
+    try {
+      result = src.rlp.decode(a)
+    } catch (e) {}
+    assert.equal(res, undefined)
+  })
+})
```
holgerd77: Ok.

For my understanding, assume some valid reasoning for it: the error cases can't be addressed with `assert.throws`?

test/externals.spec.ts
```diff
@@ -1,29 +1,167 @@
...
+      src.secp256k1.publicKeyCombine([])
+    })
+
+    // privateKeyImport invalid format
+    assert.throws(() => {
+      const buffer = Buffer.from([0x00])
+      src.secp256k1.privateKeyImport(buffer)
+    })
+  })
+})
```
holgerd77: Ok.

test/externals.spec.ts
```diff
@@ -1,29 +1,167 @@
...
+    const buf = (src as any).intToBuffer(j)
+    assert.equal(buf.toString('hex'), '5b9ac8')
+  })
+
+  it('should handle exceptions and invalid inputs', () => {
+    // should throw when invalid abi
+    assert.throws(() => (src as any).getKeys([], 3289), Error)
+
+    // should detect invalid length hex string
+    assert.equal((src as any).isHexString('0x0', 2), false)
```
holgerd77: Ok.

PR #236, opened 2020-02-05
Examples in README

### LINKED ISSUE #233: Examples in README for different categories
This library would be easier to digest respectively friendlier to be approached if there would be a comprehensive dedicated `EXAMPLES` section in the [README](https://github.com/ethereumjs/ethereumjs-util#synopsis), this can eventually be accompanied by having a look over the `README` file, there seems to be some syntax errors in the head section, some sections could also need some rewording or additional notes.

Main thing though would be to add one example code section (@ryanio: if you take on this and if this works out for the docs: this could be combined with linking to the respective API doc section) for `account`, `bytes`, eventually `constants` (why not?), `hash` and `signature`. `object` with the to-be-deprecated `defineProperties` functionality should be left out.

Additional note: I now labelled this `nice-to-have`, but actually couldn't decide on first round. I DO find these kind of updates of some significant importance, since stuff like that actually allows many people on a first encounter to use a library AT ALL and not being driven away again by getting lost through not knowing where to actually start.

ryanio:
This PR gives the README a healthy refresh along with updated examples and links to the new docs.

Closes #233.


ryanio:
@holgerd77 let me know if you'd like for me to add more examples.

PR #241, opened 2020-04-23
Enforce hex prefixing for address strings

cgewecke:
Part of #172 [Breaking API Changes][2]

The proposal to make hex param checking more strict in #172 comes out of review discussions in #170. [There][1] @holger suggested:

> Then we should at least pass input params to a dedicated check method and throw if no hex-prefix provided, and always return prefixed hex strings.

PR adds logic to enforce hex-prefixing for all address string inputs. Affected methods are:

+ isValidAddress
+ isZeroAddress
+ toChecksumAddress
+ isValidChecksumAddress  (via isValidAddress)

`isPrecompile` is excluded because it's targeted for removal in #242. 

[1]: https://github.com/ethereumjs/ethereumjs-util/pull/170#issuecomment-452318028
[2]: https://github.com/ethereumjs/ethereumjs-util/issues/172#issuecomment-491756771

### DIFF
--- a/src/account.ts
+++ b/src/account.ts
@@ -2,8 +2,9 @@ const ethjsUtil = require('ethjs-util')
 import * as assert from 'assert'
 import * as secp256k1 from 'secp256k1'
 import * as BN from 'bn.js'
-import { toBuffer, addHexPrefix, zeros, bufferToHex, unpad } from './bytes'
+import { toBuffer, zeros, bufferToHex, unpad } from './bytes'
 import { keccak, keccak256, rlphash } from './hash'
+import { assertIsHexString } from './helpers'
 
 /**
  * Returns a zero address.
@@ -17,16 +18,18 @@ export const zeroAddress = function(): string {
 /**
  * Checks if the address is a valid. Accepts checksummed addresses too.
  */
-export const isValidAddress = function(address: string): boolean {
-  return /^0x[0-9a-fA-F]{40}$/.test(address)
+export const isValidAddress = function(hexAddress: string): boolean {
+  assertIsHexString(hexAddress)
+  return /^0x[0-9a-fA-F]{40}$/.test(hexAddress)
 }
 
 /**
  * Checks if a given address is a zero address.
  */
-export const isZeroAddress = function(address: string): boolean {
+export const isZeroAddress = function(hexAddress: string): boolean {
+  assertIsHexString(hexAddress)
   const zeroAddr = zeroAddress()
-  return zeroAddr === addHexPrefix(address)
+  return zeroAddr === hexAddress
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/account.ts
```diff
@@ -115,7 +118,10 @@ export const generateAddress2 = function(
  * Returns true if the supplied address belongs to a precompiled account (Byzantium).
  */
 export const isPrecompiled = function(address: Buffer | string): boolean {
-  const a = unpad(address)
+  if (typeof address === 'string') {
+    enforceHexString(address)
+  }
+  const a = ethjsUtil.stripHexPrefix(unpad(address))
   return a.length === 1 && a[0] >= 1 && a[0] <= 8
 }
```
cgewecke, line 126: Just for reference wrt the safety of this change. Complete [stripHexPrefix][1] method is:
```javascript
module.exports = function stripHexPrefix(str) {
  if (typeof str !== 'string') {
    return str;
  }

  return isHexPrefixed(str) ? str.slice(2) : str;
}
```
[1]: https://github.com/SilentCicero/strip-hex-prefix/blob/master/src/index.js

src/account.ts
```diff
@@ -190,3 +196,14 @@ export const importPublic = function(publicKey: Buffer): Buffer {
...
+/**
+ * Helper method which throws if a string is not hex prefixed
+ * @param {string} input string to check hex prefix of
+ */
+const enforceHexString = function(input: string): void {
+  const msg = `This method only supports 0x-prefixed hex strings but input was: ${input}`
+  if (!ethjsUtil.isHexString(input)) {
+    throw new Error(msg)
+  }
+}
```
cgewecke, line 209: Using `isHexString` instead of `isHexPrefixed` here because the latter throws it's own error if given a non-string input.

src/account.ts
cgewecke: This is irrelevant.

holgerd77:
@cgewecke Great, thanks for picking this up! 😄 Did I mention that you should really question what's written up in the [API Change Table](https://github.com/ethereumjs/ethereumjs-util/issues/172#issuecomment-492118336) and give this also your own thoughts respectively reflect if you see these as useful/adequate API changes.

I've written this up quite some time ago and there actually was not yet much discussion around this yet, so I would encourage all others to give this some thought as well.

test/account.spec.ts
```diff
@@ -494,10 +510,27 @@ describe('.isValidAddress()', function() {
...
+  })
+  it('should throw when input is not hex prefixed', function() {
+    assert.throws(function() {
+      isValidAddress('2f015c60e0be116b1f0cd534704db9c92118fb6a')
+    })
+    assert.throws(function() {
+      isValidAddress('x2f015c60e0be116b1f0cd534704db9c92118fb6a')
+    })
+    assert.throws(function() {
+      isValidAddress('0X52908400098527886E0F7030069857D2E4169EE7')
```
holgerd77, line 524: Nice. 👍

PR #244, opened 2020-04-24
Make bytes methods type strict

cgewecke:
Part of #172 [Breaking API changes][1]

+ Enforces Buffer input for setLengthLeft, setLengthRight
+ Removes setLength (alias for setLengthLeft)
+ Removes stripZeros (alias for unPad)
+ Splits unpad into:
  + unpadBuffer
  + unpadHexString
  + unpadArray

(Will make similar changes for the methods in hash.ts and signature.ts in a separate PR)

[1]: https://github.com/ethereumjs/ethereumjs-util/issues/172#issuecomment-492118336

### DIFF
--- a/src/bytes.ts
+++ b/src/bytes.ts
@@ -1,6 +1,6 @@
 const ethjsUtil = require('ethjs-util')
 import * as BN from 'bn.js'
-import { assertIsBuffer } from './helpers'
+import { assertIsBuffer, assertIsArray, assertIsHexString } from './helpers'
 
 /**
  * Returns a buffer filled with 0s.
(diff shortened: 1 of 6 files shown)


src/bytes.ts
```diff
@@ -34,34 +35,62 @@ export const setLengthLeft = function(msg: any, length: number, right: boolean =
...
-  a = ethjsUtil.stripHexPrefix(a)
+const stripZeros = function(a: any): Buffer | number[] | string {
   let first = a[0]
   while (a.length > 0 && first.toString() === '0') {
     a = a.slice(1)
     first = a[0]
   }
   return a
 }
-export const stripZeros = unpad
```
cgewecke, line 64: I've removed stripZeros from the exports and made it the core logic of the unpad variants. Another approach might be to add `stripZerosBuffer`, `stripZerosHexString`, etc. Not certain what the right thing to do here is.

src/bytes.ts
```diff
@@ -10,16 +11,16 @@ export const zeros = function(bytes: number): Buffer {
...
  * Or it truncates the beginning if it exceeds.
- * @param msg the value to pad (Buffer|Array)
+ * @param msg the value to pad (Buffer)
  * @param length the number of bytes the output should be
  * @param right whether to start padding form the left or right
- * @return (Buffer|Array)
+ * @return (Buffer)
  */
-export const setLengthLeft = function(msg: any, length: number, right: boolean = false) {
+export const setLengthLeft = function(msg: Buffer, length: number, right: boolean = false) {
```
holgerd77, line 21: Hmm, this is pretty confusing to have this boolean `right` parameter in the API. Can we instead do an internal `setLength` method and move the logic over there and do function signature (so: `function(msg: Buffer, length: number)`) and internal structure analogue to `setLengthRight`?

src/bytes.ts
```diff
@@ -34,34 +35,62 @@ export const setLengthLeft = function(msg: any, length: number, right: boolean =
...
+  return setLengthLeft(msg, length, true)
 }
 
 /**
- * Trims leading zeros from a `Buffer` or an `Array`.
+ * Trims leading zeros from a `Buffer`.
+ * @param a (Buffer)
+ * @return (Buffer)
+ */
+export const unpadBuffer = function(a: any): Buffer {
```
holgerd77, line 55: I assume having an `any` type here is a mistake?

src/bytes.ts
```diff
@@ -34,34 +35,62 @@ export const setLengthLeft = function(msg: any, length: number, right: boolean =
...
+  assertIsBuffer(a)
+  return stripZeros(a) as Buffer
+}
+
+/**
+ * Trims leading zeros from a `Array` (of numbers).
```
holgerd77, line 61: Nit-pick "an `Array`"

src/bytes.ts
holgerd77: Think this is a good solution.

src/bytes.ts
cgewecke: Yes, that's definitely nicer.

src/bytes.ts
cgewecke: Ai! Thanks.

src/bytes.ts
cgewecke: Addressed in 82ab4b4

cgewecke: Addressed in 82ab4b4

src/bytes.ts
cgewecke: Addressed in 1f2803c

src/bytes.ts
cgewecke: Addressed in a2a50ae

PR #247, opened 2020-04-28
Make hash methods type strict

cgewecke:
Part of #172 [Breaking API changes][1]

+ Makes `keccak`, `sha256` and `ripemd160` buffer-only and adds `fromString` and `fromArray` variants for those inputs.

+ Adds `assertIsBuffer` check to input of `hashPersonalMessage` (excluding the possibility that someone could submit an array).

**Notes**:

There are more methods in `signature.ts` which *could* have their params validated but overall that file seems to have clearly specified input types.

In adding convenience methods, I omitted `number` because idk if that's really used. However, this has been a valid input to the hash methods. 

[1]: https://github.com/ethereumjs/ethereumjs-util/issues/172#issuecomment-492118336

### DIFF
--- a/src/hash.ts
+++ b/src/hash.ts
@@ -54,10 +54,37 @@ export const keccakFromArray = function(a: number[], bits: number = 256) {
 }
 
 /**
- * Creates SHA256 hash of the input.
- * @param a The input data (Buffer|Array|String|Number)
+ * Creates SHA256 hash of a Buffer input.
+ * @param a The input data (Buffer)
  */
-export const sha256 = function(a: any): Buffer {
+export const sha256 = function(a: Buffer): Buffer {
+  assertIsBuffer(a)
+  return _sha256(a)
+}
+
+/**
+ * Creates SHA256 hash of a string input.
+ * @param a The input data (string)
+ */
+export const sha256FromString = function(a: string): Buffer {
+  assertIsString(a)
+  return _sha256(a)
+}
+
+/**
+ * Creates SHA256 hash of a number[] input.
+ * @param a The input data (number[])
+ */
+export const sha256FromArray = function(a: number[]): Buffer {
+  assertIsArray(a)
+  return _sha256(a)
+}
+
+/**
+ * Creates SHA256 hash of an input.
+ * @param  a The input data (Buffer|Array|String)
+ */
+const _sha256 = function(a: any): Buffer {
   a = toBuffer(a)
   return createHash('sha256')
     .update(a)
--- a/src/hash.ts
+++ b/src/hash.ts
@@ -3,7 +3,7 @@ const createHash = require('create-hash')
 const ethjsUtil = require('ethjs-util')
 import * as rlp from 'rlp'
 import { toBuffer, setLengthLeft } from './bytes'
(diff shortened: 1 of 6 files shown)


src/hash.ts
```diff
@@ -3,52 +3,130 @@ const createHash = require('create-hash')
...
+ * @param bits (number = 256) The Keccak width
+ */
+export const keccakFromString = function(a: string, bits: number = 256) {
+  assertIsString(a)
+  let buf
+  if (!ethjsUtil.isHexString(a)) {
+    buf = Buffer.from(a, 'utf8')
+  } else {
+    buf = toBuffer(a)
+  }
```
holgerd77, line 41: This leads to ambiguous type interpretation situations, where it is not decidable if a string is a string randomly having a `0x` at the beginning or really being a hex string (let's take e.g. the someone wants to hash "0x is my favorite defi protocol").

Maybe sounds constructed but we had these cases (and problems with it) in the past.

First thought I had was to just take the utf8 case here and leave the hex string case to the buffer version of the method and require users to manually convert.

Second thought, which I would actually prefer since it's a common use case: would suggest to do a separate method `keccakFromHexString` with an `assertHexSrring` check and leave this method for the utf8 case. 

Or do you have a strong opinion here in another direction?

src/hash.ts
```diff
@@ -3,52 +3,130 @@ const createHash = require('create-hash')
...
+
+/**
+ * Creates RIPEMD160 hash of a string input.
+ * @param a The input data (String)
+ * @param padded Whether it should be padded to 256 bits or not
+ */
+export const ripemd160FromString = function(a: string, padded: boolean): Buffer {
+  assertIsString(a)
+  return _ripemd160(a, padded)
+}
```
holgerd77, line 112: I have a tendency not to introduce the hex string version methods for `sha256` and `ripemd160` since both are not so widely used, did some org-wide search here as well (not sure if this is a valid argument).

If you feel otherwise (that hex methods should be introduced) please do so.

test/hash.spec.ts
```diff
@@ -1,39 +1,85 @@
...
+  keccakFromArray,
+  sha256,
+  sha256FromString,
+  sha256FromArray,
+  ripemd160,
+  ripemd160FromString,
+  ripemd160FromArray,
+  rlphash,
+  toBuffer,
+} from '../src'
```
holgerd77, line 15: Likely not worth mentioning but just in case this gets forgotten: if you add the hex string methods there should be additional test cases for the new methods.

src/hash.ts
cgewecke: I think your approach is right. This is consistent with the recent changes as well. Will update.

src/hash.ts
cgewecke: That sounds good to me - really just did this for the sake of symmetry. Will update.

src/hash.ts
cgewecke: dbda438531f6e0d861db30a3efb1ea17b43efa8e

test/hash.spec.ts
cgewecke: dbda438531f6e0d861db30a3efb1ea17b43efa8e

PR #253, opened 2020-05-25
New release v7.0.2

holgerd77:
I hate to do this a bit, but I think it's best for the community in the longer run (though maybe not for our short-term release reputation 🤨 😛) to do another turn here and go back up again to `BN.js` `v5` now with the interoperability patches out.

This will ensure that the libraries will slowly get up to the latest version and will further profit from patches and upgrades (which will likely get published on the latest `v5` branch).


holgerd77:
Just published as [v7.0.2](https://github.com/ethereumjs/ethereumjs-util/releases/tag/v7.0.2) on npm.

PR #258, opened 2020-06-16
V5 Removing native js dependencies

nebojsa94:
This PR removes keccak and secp256k1 dependencies and instead uses ethereum-cryptography package that doesn't require native dependency compiling.

PR #257, opened 2020-06-10
Removing native js dependency

nebojsa94:
This PR removes keccak and secp256k1 dependencies and instead uses ethereum-cryptography package that doesn't require native dependency compiling

PR #259, opened 2020-06-16
V6 Removing native js dependencies

nebojsa94:
This PR removes keccak and secp256k1 dependencies and instead uses ethereum-cryptography package that doesn't require native dependency compiling.


src/account.ts
```diff
@@ -1,6 +1,11 @@
 const assert = require('assert')
+const {
+  privateKeyVerify,
+  publicKeyCreate,
+  publicKeyVerify,
+  publicKeyConvert,
+} = require('ethereum-cryptography/shims/hdkey-secp256k1v3')
```
alcuadrado, line 7: Why do we import these shims? Those shouldn't be considered public and will be removed once `hdkey` gets updated.

src/hash.ts
```diff
@@ -19,9 +19,23 @@ export const keccak = function(a: any, bits: number = 256): Buffer {
...
+      return k256(a)
+    }
+    case 384: {
+      return keccak384(a)
+    }
+    case 512: {
+      return keccak512(a)
+    }
+    default: {
+      throw new Error(`Invald algorithm: keccak${bits}`)
```
alcuadrado, line 36: Note for reviewers: There aren't versions of keccak other than 224, 256, 384, and 512. This API is confusing, but we have to keep it for backward compatibility reasons.

src/signature.ts
```diff
@@ -1,4 +1,5 @@
-const secp256k1 = require('secp256k1')
+const { sign, publicKeyConvert } = require('ethereum-cryptography/shims/hdkey-secp256k1v3')
```
alcuadrado, line 1: Same here

src/index.ts
```diff
@@ -1,4 +1,4 @@
-const secp256k1 = require('secp256k1')
+const secp256k1 = require('ethereum-cryptography/secp256k1')
```
alcuadrado, line 1: This variable gets exported in the line 20, so it's a breaking change. 

We need a wrapper of secp256k1 v4 that offers the same API than secp256k1 v3. 

The same happens in the version 5 of this library. Luckily, v7 doesn't reexport secp256k1 anymore.

alcuadrado: I think @nebojsa94 is working on this. Can you confirm?

holgerd77:
@nebojsa94 Thanks for the various PRs on the different EthereumJS libraries. 😄 

Can you please clarify here how this PR relates to #258 and also settle the state of #258 itself?

src/index.ts
nebojsa94: Yep

src/account.ts
```diff
@@ -123,7 +128,11 @@ export const isPrecompiled = function(address: Buffer | string): boolean {
  * Checks if the private key satisfies the rules of the curve secp256k1.
  */
 export const isValidPrivate = function(privateKey: Buffer): boolean {
-  return secp256k1.privateKeyVerify(privateKey)
+  try {
```
alcuadrado, line 131: I'm confused about why this version adds a `try/catch` while this other doesn't: https://github.com/ethereumjs/ethereumjs-util/pull/257/files#diff-49540b967120fc524b229469e9847e68R127

nebojsa94: Secp256k1 expects provided private key to be an Uint8Array with length 32. If this condition is not met the function will throw. I think this is the only case that can cause exception, I'll remove try/catch and add this check.

alcuadrado: Oh, I see. It must be a change in secp256k1 v4

alcuadrado:
This comment is also relevant here: https://github.com/ethereumjs/ethereumjs-util/pull/261#issuecomment-656753003

nebojsa94:
Here are a couple of PRs that are testing this code:
https://github.com/Tenderly/ethereumjs-abi/pull/2
https://github.com/Tenderly/ethereumjs-vm/pull/1

PR #260, opened 2020-07-02
Fix prod tsconfig and include ethjs-util types in dist

cgewecke:
Fixes two bugs introduced in #248. 

+ restores `include` option in tsconfig.prod.json. Files are no longer written to `dist/src` and `dist/test` is no longer produced
+ renames @types/ethjs-util/index.d.ts --> index.ts. This forces the compiler to generate a `.d.ts` file for it and copy everthing into dist appropriately. (More info about this approach on [stack overflow here][1].)

[1]: https://stackoverflow.com/questions/56018167/typescript-does-not-copy-d-ts-files-to-build


holgerd77:
@cgewecke thanks for addressing this so quickly, this looks already a lot better with the correct general folder structure and the types from `ethjs-util` being exported.

When I pack and test-install this I don't get access to an ES6-style import structure though (or is this technically just not possible?) nor do I get any type information, see the following screenshot:

Bildschirmfoto 2020-07-06 um 11 41 34

Could you have a look and state what should and should not be working here? 😄 

We should also add usage information on the [docs](https://github.com/ethereumjs/ethereumjs-util#ethjs-util-methods) here along (on this PR or the subsequent release PR).

holgerd77:
@nebojsa94 @alcuadrado we need to address this here until we can release `v7.0.3`.

cgewecke:
@holgerd77 Very sorry, thanks for double-checking this. 

I *think* 173e6bc fixes. Tested locally and TS was able to resolve the types from the dist.  Have also removed the added compiler options that were papering over this problem in the tests.

> We should also add usage information

The existing example shows an import statement like this:
```ts
import { isValidChecksumAddress, unpad, BN } from 'ethereumjs-util'
```
Are you thinking of something more detailed?

src/index.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference path="./@types/ethjs-util/index.ts"/>
```
alcuadrado, line 1: I think you can just do an `import "./@types/ethjs-util/index.ts";` here. I'm not sure if this triple-slash reference would work after compilation, as `index.ts` won't exist in the build.

cgewecke: @alcuadrado 

Yeah this is weird but `tsc` auto-generates the contents the @types folder in the `dist` creating a `.d.ts` file. 
```
dist/
  @types/
    ethjs-util/
      index.d.ts
      index.js
      index.js.map
```
*And* it rewrites the triple slash reference to point to the `.d.ts` in the dist's main `index.d.ts`.

alcuadrado: 🤯

PR #261, opened 2020-07-02
V4 Removing native js dependencies 

nebojsa94:
This PR removes secp256k1 dependency and instead uses ethereum-cryptography package that doesn't require native dependency compiling.


alcuadrado:
This version should be released as `v4.5.1`.

alcuadrado:
As promised, here's an explanation of this PR, and why it is so different compared to the v7 one. 

This explanation is valid for v4, v5, and v6:

* Older versions of `ethereumjs-util` re-exports their `secp256k1` dependency. This means that its API is part of this library's API.

* `ethereum-cryptography` uses `secp256k1@v4`, as it's N-API based, but that version can't be re-exported here, otherwise, we would introduce a breaking change.

* To avoid that, a wrapper was added, that consumes `secp256k1@v4` and offers `secp256k1@v3`'s API.

* In doing so, we found that some functions are no longer available in `secp256k1@v4`. To implement those, some parts of `secp256k1@v3` were imported almost verbatim. `secp256k1@v3` already has a pure js version, so these imported parts were extracted from there.

* To test this wrapper,  `secp256k1@v3`'s tests were imported. This way we can be confident on both the wrappers and the tests. Implementing a new test suite would be too much work and would not be as effective, as it could be wrong or incomplete.

This part is only valid for v4:

* This PR also adds babel, as some parts of the new code use some javascript constructs that are not available in some of the versions of node that v4's CI uses.

nebojsa94:
Here are a couple of PRs that are testing this code:
https://github.com/Tenderly/ethereumjs-abi/pull/1
https://github.com/Tenderly/merkle-patricia-tree/pull/2

PR #269, opened 2020-07-31
Fix large string copies

jochem-brouwer:
When implementing Homestead in [this PR](https://github.com/ethereumjs/ethereumjs-vm/pull/815) I ran into an issue when trying to copy SHA3 large buffers which is caused by `assertIsBuffer`. This copies Buffers with sizes which node cannot handle even if there are no errors:

```
Error: Cannot create a string longer than 0x3fffffe7 characters
    at Buffer.toString (buffer.js:777:17)
    at Object.exports.assertIsBuffer (/Volumes/Ethereum/ethereumjs-vm/node_modules/ethereumjs-util/src/helpers.ts:19:15)
    at Object.exports.keccak (/Volumes/Ethereum/ethereumjs-vm/node_modules/ethereumjs-util/src/hash.ts:14:3)
    at Object.exports.keccak256 (/Volumes/Ethereum/ethereumjs-vm/node_modules/ethereumjs-util/src/hash.ts:39:10)
```

When you try to copy a Buffer into a string which is too large, it throws. Besides that, it does not make sense to create these messages if there are no errors (this is unnecessary copying).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - run: npm install
       - run: npm run test
 
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -5,8 +5,8 @@ import * as ethjsUtil from 'ethjs-util'
  * @param {string} input string to check hex prefix of
  */
 export const assertIsHexString = function(input: string): void {
-  const msg = `This method only supports 0x-prefixed hex strings but input was: ${input}`
   if (!ethjsUtil.isHexString(input)) {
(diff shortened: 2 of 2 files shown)


jochem-brouwer:
Check won't pass because coverage doesn't upload?

jochem-brouwer:
@evertonfraga could you take a look why Upload to Coveralls is not working in the CI? 😄

evertonfraga:
Sure, sir!

evertonfraga:
The main failure was a command trying to fetch a SHA reference from git, where it didn't exist. `Checkout` version 2 fixes that.

```##[error]Error: Command failed: git cat-file -p 3130d061cb32a5063fafc16788945bd0cce0d55b```

jochem-brouwer:
Thanks a lot @evertonfraga 😄

PR #270, opened 2020-08-01
Fix the types of the BN and RLP re-exports

### LINKED ISSUE #267: BN re-export failure
Hello! I'm not being able to use the reexported version of BN.js from TypeScript due to [how the export is defined](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/externals.ts#L7). I'm using the `7.0.3` release.

Example code:
```typescript
import { BN } from 'ethereumjs-util';

function hexStringIsZero(hexNumber: string): boolean {
  return new BN(hexNumber, 'hex').isZero();
}
```

Printed error plus suggested fix:
```
../core/src/Foo.ts:51:14 - error TS2351: This expression is not constructable.
  Type 'typeof BN' has no construct signatures.

51   return new BN(hexNumber, 'hex').isZero();
                ~~

  ../core/node_modules/ethereumjs-util/dist/externals.d.ts:6:1
    6 import * as BN from 'bn.js';
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.
```

alcuadrado:
This PR fixes #267.

Here's an explanation of what I think was wrong, how I fixed, and how it can be simplified in the future.

1. BN and RLP are distributed as CJS modules, and use the `module.exports = ` pattern. This is ok in CJS, but doesn't make sense in ESM. `module.exports` is not the same as `export default`.

2. TS has a special syntax to do that, `export = obj`.

3. BN and RLP typings use that syntax.

3. To import modules using that syntax, you have to use `import A = require('A')` in ts.

4. This module was using `import * as A from 'A'` instead, which means something like "import the module namespace" in ESM. Not that module namespace is not a class in ESM, so ts gets confused if you try to initialize it. 

5. This worked anyway, as we are just reexporting it, so no error was triggered.

6. When a TS project tries to use it, ts complains about this error that went unnoticed here.

7. This make this CSJ/ESM compatibility mess easier, TS introduced the `esModuleInterop` flag, which let you use `import A as 'A'` to import CSJ modules. This breaks the ESM semantics a bit, but it's a great trade off. 

8. As I wrote in a TODO, we are gonna enable that flag in a future version of the ts config package, so this change can be simplified in the future.

### DIFF
--- a/src/externals.ts
+++ b/src/externals.ts
@@ -7,8 +7,8 @@
 // TODO: This can be replaced with a normal ESM import once
 // the new major version of the typescript config package
 // is released and adopted here.
-import BN = require('bn.js');
-import rlp = require('rlp');
+import BN = require('bn.js')
+import rlp = require('rlp')
 
 /**
  * [`BN`](https://github.com/indutny/bn.js)
--- a/src/externals.ts
+++ b/src/externals.ts
@@ -4,8 +4,11 @@
  * @packageDocumentation
  */
 
-import * as BN from 'bn.js'
-import * as rlp from 'rlp'
+// TODO: This can be replaced with a normal ESM import once
+// the new major version of the typescript config package
+// is released and adopted here.
+import BN = require('bn.js');
+import rlp = require('rlp');
 
 /**
  * [`BN`](https://github.com/indutny/bn.js)


ryanio:
thanks for the explanation :)

> As I wrote in a TODO, we are gonna enable that flag in a future version of the ts config package, so this change can be simplified in the future.

could we just add `esModuleInterop` to this repo's tsconfig and keep the export the same? that would skip the need for adding a future TODO. or is it better to use this import/require syntax

alcuadrado:
I tried your suggestion. It didn't work. I'm not sure why tbh

PR #275, opened 2020-09-25
Add Account class

ryanio:
This PR adds a new `Account` class intended as a modern replacement for `ethereumjs-account`. It has a shape of `Account(nonce?: BN, balance?: BN, stateRoot?: Buffer, codeHash?: Buffer)`.

**Instantiation**

The static factory methods assist in creating an `Account` object from varying data types: `Object: fromAccountData`, `RLP: fromRlpSerializedAccount`, and `Array: fromValuesArray`.

**Methods**: `isEmpty(): boolean`, `isContract(): boolean`, `serialize(): Buffer`

Example usage:

```typescript
import { Account, BN } from 'ethereumjs-util'

const account = new Account(
  new BN(0), // nonce, default: 0
  new BN(10).pow(new BN(18)), // balance, default: 0
  undefined, // stateRoot, default: KECCAK256_RLP (hash of RLP of null)
  undefined, // codeHash, default: KECCAK256_NULL (hash of null)
)
```

For more info see the documentation or examples of usage in `test/account.spec.ts`.


docs/README.md
```diff
@@ -6,6 +6,7 @@
 
 ### Modules
 
+* ["@types/ethjs-util/index"](modules/__types_ethjs_util_index_.md)
```
ryanio, line 9: I couldn't find a way to ignore this file in typedoc without getting the error `Could not find a declaration file for module 'ethjs-util'.`, so I guess there is no harm in having it for now, although I wish the file was filled with the type info rather than just being empty 😅

The `library` mode in typedoc is pretty new and works quite well for our purposes, so perhaps it will be fixed or improved in the future:

```
Warning: File /Users/ryanghods/dev/ethereumjs-util/src/@types/ethjs-util/index.ts
is not a module and cannot be converted in library mode
```

src/account.ts
```diff
@@ -1,23 +1,124 @@
...
+export interface AccountData {
+  nonce?: BNLike
+  balance?: BNLike
+  stateRoot?: BufferLike
+  codeHash?: BufferLike
+}
+
+type BNLike = BN | string | number
+
+type BufferLike = Buffer | TransformableToBuffer | PrefixedHexString | number
```
jochem-brouwer, line 26: I think it would be handy for future integrations to also export these types

src/account.ts
```diff
@@ -1,23 +1,124 @@
...
+   */
+  isContract(): boolean {
+    return !this.codeHash.equals(KECCAK256_NULL)
+  }
+
+  /**
+   * Returns a `Boolean` determining if the account is empty.
+   */
+  isEmpty(): boolean {
+    return this.balance.isZero() && this.nonce.isZero() && this.codeHash.equals(KECCAK256_NULL)
```
jochem-brouwer, line 120: I saw this previously in our monorepo too but forgot to raise this. I think we should also explicitly add that `this.stateRoot.equals(KECCAK256_RLP)`. 

The reason for this is that in the current HF it is not possible to instantiate a contract which would return empty (nonce is set to 1 if the tx succeeds). However, in at least Frontier nonce is not set to 1 and instead remains 0 upon creation. It is thus possible to create a contract where in the constructor you store some values in storage (thus `stateRoot` is not `KECCAK256_RLP`), but you do not put any code in. So in that case you have created a non-empty account which this function would currently say is empty. I think this might lead to some problems if we ever want to full sync mainnet.

src/account.ts
```diff
@@ -192,3 +284,21 @@ export const importPublic = function(publicKey: Buffer): Buffer {
...
+  const addressLength = 20
+  const addr = zeros(addressLength)
+  return bufferToHex(addr)
+}
+
+/**
+ * Checks if a given address is a zero address.
+ */
+export const isZeroAddress = function(hexAddress: string): boolean {
+  assertIsHexString(hexAddress)
```
jochem-brouwer, line 301: Can we explicitly check that this `hexAddress` has the expected length?

test/account.spec.ts
```diff
@@ -16,6 +18,187 @@ import {
...
+    it('should have correct codeHash', function() {
+      assert.equal(
+        account.codeHash.toString('hex'),
+        'c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470',
+      )
+    })
+  })
+
+  describe('from RLP data', function() {
+    const accountRlp = Buffer.from(
```
jochem-brouwer, line 101: Nit: we could improve this test by explicitly generating the Buffer via `rlp.encode` and then checking against the original data. From this point of view this buffer looks like a magic number 😄

test/account.spec.ts
```diff
@@ -16,6 +18,187 @@ import {
...
+      const account = new Account()
+      assert.ok(account.isEmpty())
+    })
+
+    it('should return false for a non-empty account', function() {
+      const raw = {
+        nonce: '0x01',
+        balance: '0x0042',
+        stateRoot: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421',
+        codeHash: '0xc5d2461236f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470',
```
jochem-brouwer, line 173: This test is confusing, the `codeHash` looks like `KECCAK256_NULL` yet it is not (by a few characters)

test/account.spec.ts
```diff
@@ -196,112 +379,16 @@ describe('privateToPublic', function() {
...
-      27,
-      8,
-      6,
-      126,
-      156,
-      28,
-      95,
-    ])
+    // prettier-ignore
+    const privateKey = Buffer.from([234, 84, 189, 197, 45, 22, 63, 136, 201, 58, 176, 97, 87, 130, 207, 113, 138, 46, 251, 158, 81, 167, 152, 154, 171, 27, 8, 6, 126, 156, 28, 95,])
```
jochem-brouwer, line 383: Unrelated to this PR but I think for the form it would be better to convert this to a hex string to be consistent.

test/account.spec.ts
```diff
@@ -196,112 +379,16 @@ describe('privateToPublic', function() {
...
-      8,
-      6,
-      126,
-      156,
-      28,
-    ])
+    // prettier-ignore
+    const privateKey1 = Buffer.from([234, 84, 189, 197, 45, 22, 63, 136, 201, 58, 176, 97, 87, 130, 207, 113, 138, 46, 251, 158, 81, 167, 152, 154, 171, 27, 8, 6, 126, 156, 28, 95, 42,])
+    // prettier-ignore
+    const privateKey2 = Buffer.from([234, 84, 189, 197, 45, 22, 63, 136, 201, 58, 176, 97, 87, 130, 207, 113, 138, 46, 251, 158, 81, 167, 152, 154, 171, 27, 8, 6, 126, 156, 28,])
```
jochem-brouwer, line 391: Nit: convert to hex string

test/account.spec.ts
```diff
@@ -325,40 +412,8 @@ describe('privateToAddress', function() {
...
-      27,
-      8,
-      6,
-      126,
-      156,
-      28,
-      95,
-    ])
+    // prettier-ignore
+    const privateKey = Buffer.from([234, 84, 189, 197, 45, 22, 63, 136, 201, 58, 176, 97, 87, 130, 207, 113, 138, 46, 251, 158, 81, 167, 152, 154, 171, 27, 8, 6, 126, 156, 28, 95,])
```
jochem-brouwer, line 416: Nit: convert to hex string

jochem-brouwer (commented):
Some comments regarding consistency and explicitness. Looks good in general 👍 😄

src/account.ts
ryanio: sure good idea, done in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/3f858dbc5e58e7e86dbbe365b70b48244407b2d5

src/account.ts
ryanio: good idea, added in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/cfe8a9a13d94a4d2c9ac05d2e590261f7b9a3f5b

src/account.ts
ryanio: that function wasn't introduced in this pr, i just moved it to the bottom of the file to organize.

it does check the length though in its string comparison with `zeroAddress()`?

test/account.spec.ts
ryanio: yeah we do that in testing `serialize()` a bit further down.

i copied these tests from ethereumjs-account so i'm not sure who generated the original RLP, i guess we can add a comment of the raw data, but i like one example of using a rlp hex string.

test/account.spec.ts
ryanio: done in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/6c89111e4b60b4fd3a1d98205199d6d274b61959

test/account.spec.ts
ryanio: done in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/6c89111e4b60b4fd3a1d98205199d6d274b61959

test/account.spec.ts
ryanio: done in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/6c89111e4b60b4fd3a1d98205199d6d274b61959

src/account.ts
```diff
@@ -1,23 +1,113 @@
...
+
+  /**
+   * Returns the RLP serialization of the account as a `Buffer`.
+   */
+  serialize(): Buffer {
+    return rlp.encode([bnToRlp(this.nonce), bnToRlp(this.balance), this.stateRoot, this.codeHash])
+  }
+
+  /**
+   * Returns a `Boolean` deteremining if the account is a contract.
```
holgerd77, line 94: -> determining

src/account.ts
jochem-brouwer: It does, but I do not think it is right if you input a string like "0x00" - this is not a zero address, but it is not a valid address anyways because it is too short. I'd say that raising an error here is better for the end user.

ryanio: yeah well that would return `false` which is correct, right?

i don't mind the idea of adding an error based on length, but technically that would be a breaking change since people with its current use may not expect an error to be thrown. so i'll leave it for now, but if you feel strongly, we can add in a separate pr.

test/account.spec.ts
ryanio: updated in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/36a070008a4048461d315db487cfbe9f83e16b24

src/account.ts
ryanio: thanks, fixed

src/account.ts
```diff
@@ -1,23 +1,113 @@
...
+  }
+
+  public static fromValuesArray(values: Buffer[]) {
+    const [nonce, balance, stateRoot, codeHash] = values
+
+    return new Account(
+      nonce ? new BN(nonce) : undefined,
+      balance ? new BN(balance) : undefined,
+      stateRoot,
+      codeHash,
```
holgerd77, line 59: Why do `stateRoot` and `codeHash` don't have optionality here in contrast to `nonce` and `balance`?

src/account.ts
```diff
@@ -1,23 +1,113 @@
...
+    stateRoot = KECCAK256_RLP,
+    codeHash = KECCAK256_NULL,
+  ) {
+    if (stateRoot.length !== 32) {
+      throw new Error('stateRoot must have a length of 32')
+    }
+    if (codeHash.length !== 32) {
+      throw new Error('codeHash must have a length of 32')
+    }
+
```
holgerd77, line 79: Do we want to add `nonce` and `balance` >= 0 checks here as well or is this unnecessary?

src/account.ts
ryanio: The BNs need to be wrapped in `new BN` to convert from buffer to BN. If the hashes are not provided they will default to undefined and the constructor will take care of the default values.

src/account.ts
```diff
@@ -1,23 +1,113 @@
...
+  /**
+   * Returns a `Boolean` determining if the account is a contract.
+   */
+  isContract(): boolean {
+    return !this.codeHash.equals(KECCAK256_NULL)
+  }
+
+  /**
+   * Returns a `Boolean` determining if the account is empty.
+   */
```
holgerd77, line 102: I would find it useful here to add a reference to [EIP-161](https://eips.ethereum.org/EIPS/eip-161) here for reading up on account emptiness definition, also serving as a hint where EVM behavior on emptiness of accounts was substantially changed.

src/account.ts
```diff
@@ -49,7 +130,7 @@ export const isZeroAddress = function(hexAddress: string): boolean {
  */
 export const toChecksumAddress = function(hexAddress: string, eip1191ChainId?: number): string {
```
holgerd77, line 131: Can we eventually also add an [EIP-1191](https://eips.ethereum.org/EIPS/eip-1191) link here on the respective parameter comment for convenience?

src/account.ts
```diff
@@ -192,3 +273,21 @@ export const importPublic = function(publicKey: Buffer): Buffer {
   }
   return publicKey
 }
+
+/**
+ * Returns a zero address.
+ */
```
holgerd77, line 279: Can we change "a" to "the" here for enhanced clarity?

src/account.ts
```diff
@@ -192,3 +273,21 @@ export const importPublic = function(publicKey: Buffer): Buffer {
...
+export const zeroAddress = function(): string {
+  const addressLength = 20
+  const addr = zeros(addressLength)
+  return bufferToHex(addr)
+}
+
+/**
+ * Checks if a given address is a zero address.
+ */
```
holgerd77, line 288: Same here.

src/types.ts
```diff
@@ -0,0 +1,22 @@
+import * as BN from 'bn.js'
+import { unpadBuffer } from './bytes'
+
+export type BNLike = BN | string | number
+
+export type BufferLike = Buffer | TransformableToBuffer | PrefixedHexString | number
+
+export type PrefixedHexString = string
+
+export interface TransformableToBuffer {
```
holgerd77, line 10: Can we have some code comments on the new types what makes them "BN like", "Buffer like" respectively "Transformable to Buffer"?

test/account.spec.ts
```diff
@@ -16,6 +18,187 @@ import {
...
+
+    it('should only accept an array in fromRlpSerializedAccount', function() {
+      const data = { balance: new BN(5) }
+      assert.throws(() => {
+        Account.fromRlpSerializedAccount(data as any)
+      })
+    })
+  })
+})
+
```
holgerd77, line 201: Really nice tests. 😄

holgerd77 (commented):
Somewhat comment-focused review, some suggestions for improvements, otherwise looks good.

src/account.ts
holgerd77: Ah, makes sense.

src/account.ts
ryanio: sure, added in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/b0aacc0eff2ded594a12b614508a294ad9c63f4b

src/account.ts
ryanio: added in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/e48478d3449fed6142cd6f611dffab5d4ed46ee3

src/account.ts
ryanio: fixed in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/e48478d3449fed6142cd6f611dffab5d4ed46ee3

src/types.ts
ryanio: added in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/e48478d3449fed6142cd6f611dffab5d4ed46ee3

src/account.ts
ryanio: added in fixed in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/e48478d3449fed6142cd6f611dffab5d4ed46ee3

src/account.ts
ryanio: fixed in https://github.com/ethereumjs/ethereumjs-util/pull/275/commits/e48478d3449fed6142cd6f611dffab5d4ed46ee3

holgerd77 (approved):
Thanks Ryan, looks really great! Sorry that this took so long to review, fall a bit aside, feel free to be more pushy/demanding here if you need things for subsequent tasks.

PR #277, opened 2020-09-29
Add `toBuffer()` method to Address, make `toBuffer()` function convert TransformableToBuffer

ryanio:
This PR adds a `toBuffer()` method to Address, and makes the exported `toBuffer()` function additionally handle if the object has a `toBuffer()` method.

This work was motivated by @jochem-brouwer's work in the block library refactoring [(thread here)](https://github.com/ethereumjs/ethereumjs-vm/pull/883#discussion_r496211455).

### DIFF
--- a/src/bytes.ts
+++ b/src/bytes.ts
@@ -1,5 +1,5 @@
-import * as ethjsUtil from 'ethjs-util'
 import * as BN from 'bn.js'
+import { intToBuffer, stripHexPrefix, padToEven, isHexString, isHexPrefixed } from 'ethjs-util'
 import { assertIsBuffer, assertIsArray, assertIsHexString } from './helpers'
 
 /**
@@ -86,7 +86,7 @@ export const unpadArray = function(a: number[]): number[] {
  */
 export const unpadHexString = function(a: string): string {
   assertIsHexString(a)
-  a = ethjsUtil.stripHexPrefix(a)
+  a = stripHexPrefix(a)
   return stripZeros(a) as string
 }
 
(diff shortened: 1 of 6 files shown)


test/address.spec.ts
```diff
@@ -76,4 +76,12 @@ describe('Address', () => {
       assert.equal(addr.toString(), result)
     }
   })
+
+  it('should provide a buffer that does not mutate the original address', () => {
+    const str = '0x2f015c60e0be116b1f0cd534704db9c92118fb6a'
+    const address = Address.fromString(str)
+    let addressBuf = address.toBuffer()
+    addressBuf = Buffer.from('test')
```
jochem-brouwer, line 84: This doesn't work: `Buffer.from` copies the Buffer. What we want is to actually call a method on this Buffer so it overwrites is (this basically sets `addressBuf` to some other value). For instance we can use `addressBuf.fill(0xff)`.

Example:

```
> bfr = Buffer.from('aaaaaa', 'hex')
<Buffer aa aa aa>
> bfr.fill(0)
<Buffer 00 00 00>
> bfr
<Buffer 00 00 00>
```

src/bytes.ts
```diff
@@ -105,30 +105,33 @@ const stripZeros = function(a: any): Buffer | number[] | string {
 }
 
 /**
- * Attempts to turn a value into a `Buffer`. As input it supports `Buffer`, `String`, `Number`, null/undefined, `BN` and other objects with a `toArray()` method.
+ * Attempts to turn a value into a `Buffer`.
+ * Inputs supported: `Buffer`, `String`, `Number`, null/undefined, `BN` and other objects with a `toArray()` or `toBuffer()` method.
  * @param v the value
  */
 export const toBuffer = function(v: any): Buffer {
```
jochem-brouwer, line 112: For completeness, we should set the type of `v` from `any` to all the imported inputs.

test/address.spec.ts
ryanio: oh nice thanks for explaining, that makes sense. fixed in https://github.com/ethereumjs/ethereumjs-util/pull/277/commits/18b9deadaa4399d4125c42538e7254910117cb45

src/bytes.ts
ryanio: good idea, done in https://github.com/ethereumjs/ethereumjs-util/pull/277/commits/e7aa906eb8ef53f5ec0c2581bafed1d69cec60d5

holgerd77:
Rebased this and moved the interfaces mentioned (so: TransformableToArray and TransformableToBuffer) to `./types`. Will merge once tests pass.

src/bytes.ts
```diff
@@ -105,35 +106,62 @@ const stripZeros = function(a: any): Buffer | number[] | string {
...
+  if (v.toArray) {
+    // converts a BN to a Buffer
+    return Buffer.from(v.toArray())
+  }
+
+  if (v.toBuffer) {
+    return Buffer.from(v.toBuffer())
+  }
+
+  throw new Error('invalid type')
```
holgerd77, line 164: Love this refactored and simplified code of the `toBuffer` method, also highly valuable to have proper input types here now. 😍

holgerd77 (approved):
Looks good, will merge and prepare a subsequent release PR.

PR #287, opened 2021-01-31
Handle signatures for chainIds greater than 110

spalladino:
The signature module was assuming that the `v` recovery value fit in a single byte. With chainIds larger than 110, since `v(110) = 110 * 2 + 35 = 255`, this assumption breaks (see chainid.network for a list of chainIds with large values).

This commit changes `fromRpcSig` so the recovery `v` value is not just the 65th byte of the signaure, but all bytes after the 64th.


holgerd77:
Hi Santiago, great find, thanks for the submission! 😄  Also perfect PR with sufficient test coverage together and also perfect timing (all too good to be true, hehe 😋 ) since I was just planning to do a bugfix release on the library today.

So I will directly merge and prepare a release PR.

//cc @ryanio

holgerd77 (approved):
LGTM

PR #289, opened 2021-02-04
Upgrade to ethereumjs-config v2

ryanio:
This PR upgrades this repository to ethereumjs-config v2, switches from coveralls to codecov, and fixes new lint errors. A few functions are moved up in their respective files to satisfy the `no-use-before-define` rule. 

Closes #286.

PR #290, opened 2021-02-27
signature/address: support for high recovery and chain IDs

jochem-brouwer:
In https://github.com/ethereumjs/ethereumjs-monorepo/pull/1129 I noticed that `ecrecover` does not have support for very high recovery IDs (which happen on very high chain IDs). This PR allows to pass BNs or Buffers to ecrecover, to allow this support.

I will add a test to show that this works for high chain IDs by taking a YoloV3 transaction.

### DIFF
--- a/src/bytes.ts
+++ b/src/bytes.ts
@@ -1,6 +1,6 @@
 import BN from 'bn.js'
 import { intToBuffer, stripHexPrefix, padToEven, isHexString, isHexPrefixed } from 'ethjs-util'
-import { TransformableToArray, TransformableToBuffer } from './types'
+import { PrefixedHexString, TransformableToArray, TransformableToBuffer } from './types'
 import { assertIsBuffer, assertIsArray, assertIsHexString } from './helpers'
 
 /**
@@ -112,7 +112,7 @@ export const unpadHexString = function(a: string): string {
  */
 export const toBuffer = function(
   v:
-    | string
+    | PrefixedHexString
     | number
     | BN
     | Buffer
--- a/src/bytes.ts
+++ b/src/bytes.ts
@@ -105,24 +105,24 @@ export const unpadHexString = function(a: string): string {
   return stripZeros(a) as string
 }
 
+export type ToBufferInputTypes =
+  | PrefixedHexString
+  | number
+  | BN
+  | Buffer
+  | Uint8Array
+  | number[]
+  | TransformableToArray
+  | TransformableToBuffer
+  | null
+  | undefined
+
 /**
  * Attempts to turn a value into a `Buffer`.
  * Inputs supported: `Buffer`, `String`, `Number`, null/undefined, `BN` and other objects with a `toArray()` or `toBuffer()` method.
  * @param v the value
  */
-export const toBuffer = function(
-  v:
-    | PrefixedHexString
-    | number
-    | BN
-    | Buffer
-    | Uint8Array
-    | number[]
-    | TransformableToArray
-    | TransformableToBuffer
-    | null
-    | undefined
-): Buffer {
+export const toBuffer = function(v: ToBufferInputTypes): Buffer {
   if (v === null || v === undefined) {
     return Buffer.allocUnsafe(0)
   }
--- a/src/signature.ts
+++ b/src/signature.ts
@@ -28,17 +28,24 @@ export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: any): any {
   const r = Buffer.from(signature.slice(0, 32))
   const s = Buffer.from(signature.slice(32, 64))
 
-  if (!chainId) {
-    return { r, s, v: recovery + 27 }
-  } else {
-    const chainIdBN = toType(chainId, TypeOutput.BN)
-    const vValue = chainIdBN
-      .muln(2)
-      .addn(35)
-      .addn(recovery)
-    const v = typeof chainId === 'number' ? vValue.toNumber() : vValue.toArrayLike(Buffer)
+  if (!chainId || typeof chainId === 'number') {
+    // return legacy type ECDSASignature (deprecated in favor of ECDSASignatureBuffer to handle large chainIds)
+    if (chainId && !Number.isSafeInteger(chainId)) {
+      throw new Error(
+        'The provided number is greater than MAX_SAFE_INTEGER (please use an alternative input type)'
+      )
+    }
+    const v = chainId ? recovery + (chainId * 2 + 35) : recovery + 27
     return { r, s, v }
   }
+
+  const chainIdBN = toType(chainId, TypeOutput.BN)
+  const v = chainIdBN
+    .muln(2)
+    .addn(35)
+    .addn(recovery)
+    .toArrayLike(Buffer)
+  return { r, s, v }
 }
 
 function calculateSigRecovery(v: BNLike, chainId?: BNLike): BN {
--- a/src/signature.ts
+++ b/src/signature.ts
@@ -1,9 +1,10 @@
-const { ecdsaSign, ecdsaRecover, publicKeyConvert } = require('ethereum-cryptography/secp256k1')
+import { ecdsaSign, ecdsaRecover, publicKeyConvert } from 'ethereum-cryptography/secp256k1'
 import BN from 'bn.js'
 import { toBuffer, setLengthLeft, bufferToHex, bufferToInt } from './bytes'
 import { keccak } from './hash'
 import { assertIsBuffer } from './helpers'
-import { BNLike } from './types'
+import { BNLike, PrefixedHexString } from './types'
+import { isHexString } from '.'
 
 export interface ECDSASignature {
   v: number
@@ -11,21 +12,87 @@ export interface ECDSASignature {
   s: Buffer
 }
 
+export interface ECDSASignatureBN {
+  v: BN
+  r: Buffer
+  s: Buffer
+}
+
+export interface ECDSASignatureBuffer {
+  v: Buffer
+  r: Buffer
+  s: Buffer
+}
+
+export interface ECDSASignatureHexString {
+  v: PrefixedHexString
+  r: Buffer
+  s: Buffer
+}
+
 /**
  * Returns the ECDSA signature of a message hash.
  */
-export const ecsign = function(
+export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId?: number): ECDSASignature
+export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: BN): ECDSASignatureBN
+export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: Buffer): ECDSASignatureBuffer
+export function ecsign(
   msgHash: Buffer,
   privateKey: Buffer,
-  chainId?: number
-): ECDSASignature {
+  chainId: PrefixedHexString
+): ECDSASignatureHexString
+export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: any): any {
   const sig = ecdsaSign(msgHash, privateKey)
   const recovery: number = sig.recid
 
-  const ret = {
-    r: Buffer.from(sig.signature.slice(0, 32)),
-    s: Buffer.from(sig.signature.slice(32, 64)),
-    v: chainId ? recovery + (chainId * 2 + 35) : recovery + 27
+  let ret
+  if (typeof chainId === 'number') {
+    return {
+      r: Buffer.from(sig.signature.slice(0, 32)),
+      s: Buffer.from(sig.signature.slice(32, 64)),
+      v: recovery + (chainId * 2 + 35)
+    }
+  } else if (BN.isBN(chainId)) {
+    ret = {
+      r: Buffer.from(sig.signature.slice(0, 32)),
+      s: Buffer.from(sig.signature.slice(32, 64)),
+      v: (chainId as BN)
+        .muln(2)
+        .addn(35)
+        .addn(recovery)
+    }
+  } else if (Buffer.isBuffer(chainId)) {
+    ret = {
+      r: Buffer.from(sig.signature.slice(0, 32)),
+      s: Buffer.from(sig.signature.slice(32, 64)),
+      v: toBuffer(
+        new BN(chainId)
+          .muln(2)
+          .addn(35)
+          .addn(recovery)
+      )
+    }
+  } else if (typeof chainId === 'string') {
+    if (!isHexString(chainId)) {
+      throw new Error(`A chainId string must be provided with a 0x-prefix, given: ${chainId}`)
+    }
+    ret = {
+      r: Buffer.from(sig.signature.slice(0, 32)),
+      s: Buffer.from(sig.signature.slice(32, 64)),
+      v:
+        '0x' +
+        new BN(toBuffer(chainId))
+          .muln(2)
+          .addn(35)
+          .addn(recovery)
+          .toString('hex')
+    }
+  } else {
+    ret = {
+      r: Buffer.from(sig.signature.slice(0, 32)),
+      s: Buffer.from(sig.signature.slice(32, 64)),
+      v: recovery + 27
+    }
   }
 
   return ret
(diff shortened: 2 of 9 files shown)


src/signature.ts
```diff
@@ -44,17 +47,17 @@ function isValidSigRecovery(recovery: number): boolean {
...
-  chainId?: number
+  chainId?: number | BN | Buffer
 ): Buffer {
   const signature = Buffer.concat([setLengthLeft(r, 32), setLengthLeft(s, 32)], 64)
   const recovery = calculateSigRecovery(v, chainId)
   if (!isValidSigRecovery(recovery)) {
     throw new Error('Invalid signature v value')
   }
-  const senderPubKey = ecdsaRecover(signature, recovery, msgHash)
+  const senderPubKey = ecdsaRecover(signature, recovery.toNumber(), msgHash)
```
jochem-brouwer, line 60: Note: recovery is a number between 0-3, see https://github.com/cryptocoinjs/secp256k1-node/blob/024fbdad3fb64499e8db8b35246c2f0a36afa8c8/lib/index.js#L294 (so we are fine to convert this to a number). I was a bit afraid that we would have to open an issue at secp256k1 to add BN/Buffer support for recovery ids.

alcuadrado:
Really curious about this PR. Is this because of EIP155? Some chainIDs don't fit in a number, so computing the V requires a big number?

jochem-brouwer:
@alcuadrado Yep, this was a problem on the YoloV3 network, which has a chain id of `34180983699157880`. If we don't use the logic as in this PR, then at some point when syncing, we recover the wrong sender address (probably because some rounding errors in the arithmetic). 

I will add this transaction as test.

alcuadrado:
You know what's extra curious about it? In most platforms that's not a BN! It fits in a uint64, but js 🥲

jochem-brouwer:
Okay, isolated the failing transaction on Yolo and added it as test. Ready for review.

src/signature.ts
holgerd77: Do we then eventually want to use the [BNLike](https://github.com/ethereumjs/ethereumjs-util/blob/master/src/types.ts#L8) type to make these type of input formats consistent throughout the library (also for further additions)?

Along the same line:

1. Any reason to not add `Buffer` to the `BNLike` type, then we encompass all the functionality from here? 🤔 

2. Also: Would this be a candidate for a first `BigInt` introduction on the libraries, and so: can we also add a `BigInt` input type to `BNLike`? Or does this have side effects?

jochem-brouwer: I'd say that `BNLike` is of type `Buffer | BN | number | string | BigInt`. We can first start by expanding it with `Buffer`, and then on a seperate PR include `BigInt`s. I think it is very helpful if we add this flexible support in the libraries, it is especially helpful for our consumers. If they want a simple example, it doesn't make a lot of sense to convert the number `1` to a `BN`: this adds extra logic and an extra import. On the other hand, we do make sure our users are very strict and don't run into these rounding errors. But I would be very much in favor to add `BNLike` as argument type to some arguments, and this would definitely be the first library to start with.

Another side note: we might want to introduce this package to our monorepo. ( 😅  )

holgerd77:
Yes, also thought after writing that we likely should leave `BigInt` out for now and extract to a separate issue.

Not sure, I know we want to get this released soon since the client release depends on this, on the other hand we might want to also adopt the other methods as well who are touched by this since this would better justify a release and I think this is something others would directly profit from.

Have written together a short list to get an overview on the task: 

- Account: `toChecksumAddress = function(hexAddress: string, eip1191ChainId?: number)`
  - -> relatively easy, only string conversion, would benefit, just tested a larget string output: '2.3409328409382408e+42' 😜
- Account: `isValidChecksumAddress = function( hexAddress: string, eip1191ChainId?: number)
  - -> simple, just calls into `toChecksumAddress`
- Signature: `ecsign = function(msgHash: Buffer, privateKey: Buffer, chainId?: number)`
  - -> Can we do this with function overloading to not make this breaking (return with BN (or alternatively: always Buffer? when `chainId` is `BN`) (a bit tricky)
- Signature: `toRpcSig = function(v: number, r: Buffer, s: Buffer, chainId?: number): string` (`v` and `chainId`)
  - -> needs some function overloading in the subcalls
- isValidSignature = function(v: number, r: Buffer, s: Buffer, homesteadOrLater: boolean = true, chainId?: number) (`v` and `chainId`)
  - -> needs some function overloading in the subcalls (redundant)
  
  I think this should be doable, WDYT? 🤔  If you want you can directly integrate here but I could also do tomorrow if you don't find the extra time.

For the function overloading cases I guess the most clean way of doing this is likely to always return the same type as the input type? This would then be slightly more work, but I guess it's better than always returning a `BN` when people are e.g. working in a Buffer context.

Monorepo integration: @ryanio has also already suggested this and therefore already did #289 in preparation. 😋  I don't have the capacity right now to do the integration and I also feel that the monorepo is not enough on stable ground right now to do such a heavy integration on top (CI/Coverage often failing and things). But this is definitely planned! 😄

holgerd77:
(or @ryanio you can also directly pick this up if you want to! 😄 )

jochem-brouwer:
Yeah this should not be so hard. So what is the plan here? For anything where it would make sense to allow `Buffer/BN/string/number` input we allow these? How do we internally want to handle this? I assume we will convert them all to a `BN` or a `Buffer`, depending on the situation? 

In future PRs we can start to use `BigInt` or `Uint8Arrays` (as opposed to Buffer, for optimization)

src/signature.ts
ryanio: Handling `BNLike` would be nice from a dev ux perspective, like we do in the `fromXData` factories in our monorepo packages. Maybe we can make some kind of utility/formatter to consume `BNLike` and output a specific type so it can be reused in other places to simplify the conversion logic.

> Another side note: we might want to introduce this package to our monorepo. ( 😅 )

yes I think it's a good time to do this, I have upgraded the config packages to v2 recently so @holgerd77 let me know if there's any other prep work for this repo before it would be eligible for merging into the monorepo.

ryanio:
Handling `BNLike` would be nice from a dev ux perspective, like we do in the `fromXData` factories in our monorepo packages. Maybe we can write some kind of utility/formatter to consume `BNLike` and output a specific type (`BN` or `Buffer`) so it can be reused in other places.

holgerd77:
> Yeah this should not be so hard. So what is the plan here? For anything where it would make sense to allow `Buffer/BN/string/number` input we allow these? How do we internally want to handle this? I assume we will convert them all to a `BN` or a `Buffer`, depending on the situation?

I think it makes sense to on this first round stay somewhat narrow here and not do *all* methods somehow accepting numbers, so e.g. for many of the buffer-related methods (`setLengthLeft()`,...) I would doubt that we need `BN`s there, or at least that would likely need some discussion. But on a first look I would say that these `v` respectively `chainId` related methods are safe candidates to do. So I would suggest we stick to the (I think complete in this regard) list from above.

Yes, I think in most cases, an internal conversion to `BN` should work well - e.g. for `toChecksumAddress` and similar methods. Not completely sure about the overload methods - e.g. `ecsign` - might be most practical to stay in the respective type if we choose a `BN` in-and-out, `Buffer`-in-and-out,... methodology, but this will likely becoming more clear when working on this.

holgerd77:
Phew, this is a lot harder than I expected. Already had various iterations on this, I started by expanding the `ecsign` method. Will nevertheless continue to work here since I do think that we need this anyhow if we e.g. want to expand on PoA block signing and the like. And this is generally useful to have as a feature.

1. I first added the `BNLike` type both as an input type as well as a return type. This is not so beautiful since it reduces on type safety for the return type and also directly breaks typescript code which assumed before to get a number back
2. Then I switched to function overloads - I pushed the state of this work within 5c80ce8 so you can get an impression. This *does* work, tests also pass, but when having a somewhat distant look again I am thinking that this code is getting too complex and error prone for what it does
3. Then thought: maybe just always return a `BN`: but this is not so beautiful from a future `BigInt` support perspective
4. So I finally came to the conclusion (long way), also when looking at the `ecrecover` method: the right thing to do here
  - Add just one additional interface `ECDSASignatureBuffer`
  - Use this as a return type for all inputs except the existing ones (undefined, number) (and therefore generally return a Buffer - as in `ecrecover` - which can be easily converted to `BN` or `BigInt
  -  Mark the "return as number" as well as `ECDSASignatureBuffer` (this will be renamed to `ECDSASignatureBuffer`) as deprecated and switch to always returning a Buffer on the next breaking release
  
  Hmpfg.

src/signature.ts
```diff
@@ -1,41 +1,79 @@
-const { ecdsaSign, ecdsaRecover, publicKeyConvert } = require('ethereum-cryptography/secp256k1')
+import { ecdsaSign, ecdsaRecover, publicKeyConvert } from 'ethereum-cryptography/secp256k1'
```
holgerd77, line 1: There was no specific reason that we had done this `require` here, right? 🤔

holgerd77:
Ok, this 24279bf839eb697a8f8e2b643662f6d2986c266b is finally how I would keep it for this specific function, will continue with the other methods today or tomorrow.

holgerd77:
Have added a planned-breaking-change note to https://github.com/ethereumjs/ethereumjs-util/issues/291

holgerd77:
Ok, this is now ready for review. 😄 

Have left out `fromRpcSig()` - from the doc comments I have the impression that this method is outdated and we can deprecate? Not sure about `toRpcSig()`, have adopted for now.

Otherwise I think the main v and chainId related functionality should be covered.

Side note: first planned to also include a release commit to get this out quicker (PR is needed for continued client work), but since this has now gotten so large I've omitted to not overload.

holgerd77:
(or does someone feel uneasy here with opening to so many type inputs? we can alternatively also just limit to (the existing) number + Buffer)

holgerd77:
(wonder since we have lately rather reduced type variety)

holgerd77:
(I have *some* (but no strong) tendency to think that in this case it is ok)

src/signature.ts
ryanio: seems fine to me

src/bytes.ts
```diff
@@ -112,7 +112,7 @@ export const unpadHexString = function(a: string): string {
  */
 export const toBuffer = function(
   v:
-    | string
+    | PrefixedHexString
```
jochem-brouwer, line 115: I'm assuming that originally we expected these strings to be `0x` prefixed?

ryanio: yeah actually `PrefixedHexString` is simply a string type right now, maybe in the future typescript will give us some way to enforce that the first two characters are `0x`.

src/signature.ts
```diff
@@ -1,31 +1,98 @@
...
+): ECDSASignatureHexString
+export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: any): any {
   const sig = ecdsaSign(msgHash, privateKey)
   const recovery: number = sig.recid
 
-  const ret = {
-    r: Buffer.from(sig.signature.slice(0, 32)),
-    s: Buffer.from(sig.signature.slice(32, 64)),
-    v: chainId ? recovery + (chainId * 2 + 35) : recovery + 27
+  let ret
```
jochem-brouwer, line 34: Can we explicitly add types here to `ret` so we know in which `if` block we are using which return value? So ret would be of type `ECDSASignature | ECDSASignatureBN | ECDSASignatureBuffer` (possibly have to add `undefined` to keep TypeScript happy).

ryanio:
looks great, however instead of converting the types inside each method (seems repetitive) what if we introduced a helper method `toType(from: any, to: TypeOutput)` where `TypeOutput` is an enum of `[Number, Buffer, BN]` (in the future can add `BigInt`)

Then we can also refactor out `vAndChainIdTypeChecks` and include the 0x-prefix and MAX_SAFE_INTEGER checks inside `toType` if it is given a number or string.

jochem-brouwer (commented):
Few comments here.

Fun fact, I cannot approve my own PR 😛

holgerd77:
Hi @ryanio can you eventually finish this PR? I won't be able to finish today or tomorrow and would be nice if we get this done, I'm fine with all changes. 😀

ryanio:
> Hi @ryanio can you eventually finish this PR? I won't be able to finish today or tomorrow and would be nice if we get this done, I'm fine with all changes. 😀

yes for sure I can do in a few hours when I'm back on my computer 👍

holgerd77:
(so finish === include eventual improvement suggestions, otherwise this should be mostly ready)

ryanio:
ok I added a commit (https://github.com/ethereumjs/ethereumjs-util/pull/290/commits/243fd4776d789c7b8c37dc22d0249465d1a0766c) that adds a `toType` helper that includes the 0x-prefix and MAX_SAFE_INTEGER checks so I was able to reduce some of that duplicate code, let me know what you guys think.

I would appreciate some TypeScript guru help, I was able to get the `toType` return type based on the `outputType` enum with some fancy code [here](https://github.com/ethereumjs/ethereumjs-util/pull/290/commits/243fd4776d789c7b8c37dc22d0249465d1a0766c#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR89-R92), however I wasn't able to get the return values inside the method to work seamlessly (I had to add `as any` to all the returns). Otherwise when using the `toType` method in the actual code as a helper it seems to be reporting the correct return type so that's exciting!

I might also want to add some independent tests for `toType` if we will be exporting it for use by consumers of the package, but I can do that in a follow-up PR if we want to merge this in.

(BTW @holgerd77, do you think it was okay to refactor out the "Performance optimization" [here](https://github.com/ethereumjs/ethereumjs-util/pull/290/commits/243fd4776d789c7b8c37dc22d0249465d1a0766c#diff-ae151444b91b7e5e0a6aa753024f7900f0ff0447afaa2d18ad6619d5c3e12318L149)? I can add it back in, but opted for the new simplicity and built in checks introduced by the `toType` helper.)

src/signature.ts
ryanio: refactored the code here, should be good now, please feel free to comment if it could still be improved

ryanio (approved):
this lgtm now! I would appreciate another set of :eyes: on my latest commits but otherwise  :thumbsup:

src/signature.ts
```diff
@@ -22,79 +21,40 @@ export interface ECDSASignatureBuffer {
  * Returns the ECDSA signature of a message hash.
  */
 export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId?: number): ECDSASignature
-export function ecsign(
-  msgHash: Buffer,
-  privateKey: Buffer,
-  chainId: BN | string | Buffer
-): ECDSASignatureBuffer
+export function ecsign(msgHash: Buffer, privateKey: Buffer, chainId: BNLike): ECDSASignatureBuffer
```
holgerd77, line 24: This was intentional to split the `BNLike` type to these two overlays and not use the `BNLike` type here. Otherwise there is an ambiguity on the `number` type.

Update: I just checked in the code, I would have expected for `TypeScript` to assume `ECDSASignature | ECDSASignatureBuffer` as a return type when used with a `number` `chainId`, this is not the case though, seems `TypeScript` takes the first working signature.

Will therefore let this pass, feeling a little uneasy on this nevertheless.

holgerd77 (approved):
Hi @ryanio,
thanks for these additions, this `toType()` method is really genius and a great helper. I have to say that I feel - a bit - uneasy nevertheless to have that much code with such specific functionality and walking on relatively unproven ground somewhat last-minute to a PR.

Maybe on a next round we want to do things like this in a separate PR and in doubt keep out of an outstanding release.

Will give this a go here nevertheless on this turn, I *DO* think this should be ok after having checked the code twice. Also checked that the types are working when `toType` is applied (was also somewhat irritated by these `as any` casts at first), so that's great! 😄 

Ah, and refactoring out the performance optimization seems to be ok for me, thought this might be faster, but had no real insight here and wast just a precautious measure, eventually overblown.

Ok, will directly prepare a release here. 😄

PR #295, opened 2021-06-16
Add EIP-2098 support to work with compact 64-byte signatures

k06a:
https://eips.ethereum.org/EIPS/eip-2098

### DIFF
--- a/src/signature.ts
+++ b/src/signature.ts
@@ -97,26 +97,60 @@ export const toRpcSig = function(v: BNLike, r: Buffer, s: Buffer, chainId?: BNLi
 }
 
 /**
+ * Convert signature parameters into the format of Compact Signature Representation (EIP-2098).
+ * @returns Signature
+ */
+ export const toRpcSigCompact = function(v: BNLike, r: Buffer, s: Buffer, chainId?: BNLike): string {
+  const recovery = calculateSigRecovery(v, chainId)
+  if (!isValidSigRecovery(recovery)) {
+    throw new Error('Invalid signature v value')
+  }
+
+  let vn = toType(v, TypeOutput.Number)
+  let ss = s
+  if ((vn > 28 && vn % 2 === 1) || vn === 1 || vn === 28) {
+    ss = Buffer.from(s)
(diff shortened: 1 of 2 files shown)


src/signature.ts
```diff
@@ -96,27 +96,61 @@ export const toRpcSig = function(v: BNLike, r: Buffer, s: Buffer, chainId?: BNLi
   return bufferToHex(Buffer.concat([setLengthLeft(r, 32), setLengthLeft(s, 32), toBuffer(v)]))
 }
 
+/**
+ * Convert signature parameters into the format of Compact Signature Representation (EIP-2098).
+ * @returns Signature
+ */
+ export const toRpcSigCompact = function(v: BNLike, r: Buffer, s: Buffer, chainId?: BNLike): string {
```
axic, line 103: I do not think this is related to the RPC, or does any `eth_` RPC method support this now?

If not, I'd suggest something like `toEIP2098CompactSig`, `toEIP2098Sig`, or `toCompactSig` (least preferred is the last as the proposal may not be widespread enough to have "compact sig" necessarily connect people with the EIP)

holgerd77:
@k06a hi there, cool, thanks for the PR 😄 . I could imagine that we are open for inclusion (a short pre-discussion wouldn't have been hurt either though, always some chance to get rejected which then leads to some frustration).

Short initial note: this repo is not active any more and `Util` has moved over to our monorepo https://github.com/ethereumjs/ethereumjs-monorepo. Please resubmit there for review and inclusion.

Thanks!

axic:
@holgerd77 can this repo be marked archived or there are other reasons for it to be non-archived?

holgerd77:
@axic hmm, there is a semiish reason since we still have got the v4, v5 and v6 branches here which theoretically might need some bugfix release in the future. This hasn't happened for several months though and in doubt we can then also un-archive for a short amount of time. So will do, likely cleaner and less prone for confusion.

@k06a I will generally close here, would be great if you can resubmit to https://github.com/ethereumjs/ethereumjs-monorepo

Thanks! 😄

